PR #3, opened 2020-05-06
Super basic travis CI setup.

mhvk:
With the idea that we should have some testing to ensure that whatever we have works, and that we keep a working state.

@eteq - I'm happy to set up travis but I don't see to have access to those settings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: python
+python: 3.8
+
+install:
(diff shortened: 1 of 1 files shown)


avalentino:
This PR seems to be superseded by PR #4

mhvk:
Since there seems to be agreement on the path in #4, this indeed is superfluous, so closing.

PR #4, opened 2020-05-06
Add erfa source code as a git submodule

mhvk:
The first PR with a real change, opening this for discussion: how best to include the source code? Here, I chose to use a sub-module, which seems logical as a way to have the C files simply available without having to vendor them, but I am not completely sure it is the best procedure.

With it, though, I can install purely using `pip`! This would seem important if we want to put this on `pypi`, and, shorter-term, to have a test PR in astropy that removes `_erfa` and `cextern/erfa` and instead relies on `pyerfa` (I think we may want to end up bundling `pyerfa` at first, but that's a separate discussion).

To make testing easier and compatible with what we do in astropy, I also set up a simple `tox.ini` - now `tox -e test` has passing tests! (Hopefully not relying on that I have `liberfa` installed...) I also made a `.travis.yml` to get started on that, but clearly that needs registering `pyerfa`... (and the same for doing documentation - for now, it is just an empty directory to have something for `pytest` to look at).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,32 @@
+language: python
+
+# We need a full clone to make sure setuptools_scm works properly
+git:
+    depth: false
+
+os:
+    - linux
+
+dist: bionic
+
+env:
+    global:
+        # The following three variables are for tox. TOXENV is a standard
+        # variable that tox uses to determine the environment to run,
+        # TOXARGS are arguments passed to tox, and TOXPOSARGS are arguments
+        # that tox passes through to the {posargs} indicator in tox.ini.
+        # The latter can be used for example to pass arguments to pytest.
+        - TOXARGS='-v'
+        - TOXPOSARGS=''
+
+jobs:
+    include:
+        - name: Tests that all the basics are covered.
+          python: 3.8
+          env: TOXENV=test
+
+install:
+    - pip install tox;
+
+script:
+    - tox $TOXARGS -- $TOXPOSARGS;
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -0,0 +1,32 @@
...
+        # that tox passes through to the {posargs} indicator in tox.ini.
+        # The latter can be used for example to pass arguments to pytest.
+        - TOXARGS='-v'
+        - TOXPOSARGS=''
+
+jobs:
+    include:
+        - name: Tests that all the basics are covered.
+          python: 3.8
+          env: TOXENV=test
```
astrofrog, line 26: Maybe add the oldestdeps and devdeps jobs here too? Initially we should maybe add the numpy dev *not* as a cron job, but just move it to a cron job if we start to see failures that we can't fix quickly?

astrofrog: Also, maybe worth adding a MacOS X and Windows job? (see the core package for how to do that as it requires conda for setting up Python correctly).

mhvk:
@astrofrog - agreed that we should add more jobs! Was thinking of doing that in a follow-up PR, once travis is actually running on PRs.

Just to be sure: do you also think the submodule is the right way to go? If so, maybe merge this one so CI starts and then I'll create a clean-up PR that adds jobs, removes the `Makefile` (@avalentino - it was rather neat! just not sure how to integrate that with `pip`), and add some super-minimal docs run that produces the API (obviously, the latter will require setting up `readthedocs` as well).

PR #5, opened 2020-05-06
Add base docs

avalentino:
Add documentation to pyerfa.

TODO:

- [x] rebase on top of PR #4 as soon as it is accepted (the first 2 commits shall be dropped)
- [ ] ~~drop sphinx-astropy~~
- [ ] ~~switch to the ??? theme~~
- [x] check the configuration
- [ ] ~~setup RTD config~~
- [x] check if the solution with verbatim C comments is OK
- [ ] ~~possibly replace inclusion of portions of the ``README.rst`` (fragile) with more extended contents~~


README.rst
```diff
@@ -23,7 +23,7 @@ Setup the ERFA_ source code
 
 The ERFA_ source code, necessary to build the `erfa` Python_ package, is
```
mhvk, line 24: This is now no longer true...

mhvk:
@avalentino - are editing? I had a quick look and can a rebased version to your branch that also adds a travis job.  For now, I've disabled erroring on warnings. But if you are in the middle of things, I won't...

avalentino:
@mhvk  I'm going to rebase right now ...

mhvk:
@avalentino - if you haven't done it yet, then let me push up my solution - it will save you some time!

avalentino:
@mhvk thanks

astrofrog:
We might want to explicitly not use sphinx-astropy or at the very least not use the bootstrap-astropy theme since I think we want to make sure the package is broader than astropy.

mhvk:
@astrofrog - good point. Still, I guess it is good to start from a working configuration and then remove items to make it more general. I also think we may need some special treatment of the docstrings, such that no typesetting is attempted for the parts that are scraped from the C functions (or that those are interpreted correctly; likely trickier...)

avalentino:
> We might want to explicitly not use sphinx-astropy or at the very least not use the bootstrap-astropy theme since I think we want to make sure the package is broader than astropy.

@astrofrog , @mhvk ok, if you agree I will remove sphinx-astropy.
Do you have any preference for the theme: alabaster or RTD, ... or, maybe, something elese? 

>  I also think we may need some special treatment of the docstrings, such that no typesetting is attempted for the parts that are scraped from the C functions (or that those are interpreted correctly; likely trickier...)

@mhvk at the moment the original C documentation is displayed verbatim already. The result is not too bad IMHO. Do you have something different in mind?

mhvk:
@avalentino - I would use as simple a theme as possible; on the C docs - when I tried `tox -e build_docs`, it did not look like it was reproducing it verbatim locally.

More generally, I think we can do these things in follow-up PRs - I think most important now is just to have a working version, so that we have a basis to work from more incrementally. So, I'd suggest to merge this, set up readthedocs, and raise issues for the theme and (possibly) the way docstrings look.

avalentino:
>  on the C docs - when I tried tox -e build_docs, it did not look like it was reproducing it verbatim locally.

@mhvk probably you `erfa/core.py` has not been re-generated after the switch to the `docs` branch. I made a specific change in the code generator for that.

> More generally, I think we can do these things in follow-up PRs - I think most important now is just to have a working version, so that we have a basis to work from more incrementally. So, I'd suggest to merge this, set up readthedocs, and raise issues for the theme and (possibly) the way docstrings look.

ok

mhvk:
Ah, yes, you are right. Of course that means something is not quite set up right, since `tox` should regenerate those... Also, the colouring is a bit odd for that text, with python-specific words being marked green.

The other strange thing is that it looks like the docs build is not actually executed on travis...

Anyway, I think those are other issues, so let me get this in now so that we can set up readthedocs. 

Thanks so much for actually getting this move going! I was rather dreading it!

astrofrog:
@avalentino @mhvk - maybe we should get rid of the ``os.path.exists`` check here:

https://github.com/liberfa/pyerfa/blob/master/setup.py#L33

To make sure it is always regenerated?

avalentino:
@astrofrog please see my comment in https://github.com/liberfa/pyerfa/pull/13#issuecomment-625040668

PR #13, opened 2020-05-07
Ensure we always run erfa generator when creating the package.

### LINKED ISSUE #11: Package creation should ensure core.py and ufunc.c are up to date
The auto-generated `core.py` and `ufunc.c` should be up to date with their templates before a package is created. At least in one occasion, `tox -e build_docs` created pages that did not have an update propagated through.

mhvk:
Even if files are present, they may be out of date. To get this to work, we have to ensure erfa_generator is included in the manifest.

@avalentino - since you added the check to `setup.py`, do you agree it is OK to leave it out?

fixes #11

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,6 +2,7 @@ include LICENSE.rst
 include README.rst
 include pyproject.toml
 include setup.cfg
+include erfa_generator.py
 
 recursive-include erfa *.c *.h *.templ
 
--- a/setup.py
+++ b/setup.py
@@ -2,8 +2,9 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
 import os
-
+import sys
 import setuptools
+import subprocess
 
 
 ERFAPKGDIR = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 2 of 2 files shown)


avalentino:
I'm OK with the inclusion of the `erfa_generator.py` in the source package (I was sure it already was).

Regarding the unconditioned generation of `erfa/core.py` and `erfa/ufunc.c`, in principle it is OK since we are going to include the `liberfa/erfa` code in the source distribution.

Anyway it could be a little bit annoying for packagers that may want to strip the erfa C code out of the archive and use the system liberfa.

My recommendation is to include both the generate code (`core.py` and `ufunc.c`) and the liberfa C code in the source distribution but the re-generation of `core.py` and `ufunc.c` shall be conditioned to the actual presence of `liberfa/erfa/src`.

Do you think it is acceptable?

mhvk:
@avalentino - I think your logic makes sense. It may be easiest if I first add a test that uses a system-provided library; we can then adjust it to check that without the submodule, the package still works.

mhvk:
Thinking more about it, I think we do need this right now, and we already do include the relevant files in our package, so I think it can be merged. More generally, we should think how to deal with system libraries, since these do not necessarily have the same version, i.e., we need a mechanism to check.

PR #15, opened 2020-05-07
Docs theme

### LINKED ISSUE #10: Documentation from C functions should not be coloured
Right now, the C function docstring is typeset as fixed width, but still coloured as if it is python:

image

avalentino:
Use the default sphinx theme and drop sphinx-astropy.

Also closes #10.


docs/index.rst
```diff
@@ -1,3 +1,8 @@
+.. PyEPR documentation master file, created by
+   sphinx-quickstart on Wed May  6 22:02:11 2020.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
```
astrofrog, line 4: I think we can remove these comments - if you want to keep them, PyEPR -> PyERFA

avalentino: OK, will remove.
thanks

mhvk:
This looks good, and is approved, so will merge. Thanks!

PR #20, opened 2020-05-07
Revert dropping of astropy theme

### LINKED ISSUE #17: Documentation now less good...
@avalentino, @astrofrog - Oops, I should not just have merged: the documentation is now really not so nice any more, in particular the way the API is listed as an ugly very wide-spaced table and the actual python arguments (see below). Shall I revert?

image

mhvk:
fixes #17

I think the astropy theme looks substantially better, so I thought for now we should just revert rather than spent time improving. I also think there is a benefit for most of us, in that thing behave the same as what we're used to!

I did keep the switch off of high-lighting from the reverted commit (which solved #10) as well as the reorganization from the PR. With this:

image

image


docs/conf.py
```diff
@@ -102,11 +120,12 @@
...
 
 # Grouping the document tree into LaTeX files. List of tuples
-# (source start file, target name, title,
-#  author, documentclass [howto, manual, or own class]).
+# (source start file, target name, title, author, documentclass [howto/manual]).
 latex_documents = [('index', project + '.tex', project + u' Documentation',
                     author, 'manual')]
 
+latex_logo = '_static/astropy_logo.pdf'
+
```
avalentino, line 128: `latex_logo` should be removed IMHO, the `_static` folder is not in git.

mhvk: Yes, makes sense. Done.

PR #27, opened 2020-05-09
Added initial Azure pipelines configuration for auto-release

### LINKED ISSUE #8: Consider using Azure pipelines template for auto-releasing package
@Cadair and I have created an Azure pipelines template for auto-releasing Python packages, including wheels for all the different platforms:

https://openastronomy-azure-pipelines.readthedocs.io/en/latest/

With this, pushing a tag to GitHub results in the release happening automatically. Since this is a new project, it might be worth setting this up here. An alternative would be to figure out how to do this as a GitHub action, but that would require more work for now. I do want to try and do this at some point, but if we wanted something soon we could start with the Azure pipelines version.

I'd be happy to set this up if others are interested. I use this for reproject and sunpy use it for the main sunpy package, and it works great. I also want us to use something like this for the core package in the long term.

astrofrog:
The way this works is that the wheel build and testing happens for every commit and PR, but the publication to PyPI happens only if a tag is pushed to the main repository. Let me know if the release instructions are clear enough. Note that this builds both the sdist and the wheels.

See https://openastronomy-azure-pipelines.readthedocs.io/en/latest/publish.html if you are interested in the pipeline settings. I'll be setting up the authentication on Azure Pipelines shortly, but for now the CI will run (it just would fail at the publishing step if this were building a tag)

Fixes #8

### DIFF
--- a/RELEASING.rst
+++ b/RELEASING.rst
@@ -15,6 +15,13 @@ PGP keys set up. Then, push the tag to GitHub, e.g.::
 
     git push upstream v0.1
 
-and the build should happen automatically on Azure pipelines. See
-the [OpenAstronomy Azure Pipelines Templates Documentation](https://openastronomy-azure-pipelines.readthedocs.io/en/latest/publish.html)
-for more details.
+and the build should happen automatically on Azure pipelines. You can
+follow the progress of the build here:
+
+https://dev.azure.com/liberfa/pyerfa/_build
+
+If there are any failures, you can always delete the tag, fix the
+issues, tag the release again, and push the tag to GitHub.
+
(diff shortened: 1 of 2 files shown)


astrofrog:
Note that we could also not build the wheels for all PRs and so on but just have them enabled for commits to master and tags. But perhaps PRs to here are going to be rare enough that it's worth just doing the full checks each time (it's a test of different Python versions and operating systems anyway)

astrofrog:
@mhvk - the apt build has started to fail - any ideas?

astrofrog:
> Looks good! I think between this and travis we probably cover most of the parameter space. Do these use the oldest allowed numpy? If so, perhaps a numpy-dev and erfa-dev run on travis.

They *build* with the oldest allowed Numpy but test with the most recent, so we still don't have tests running with the oldest allowed Numpy.

Issue #23, opened 2020-05-09
Prepare release?


mhvk:
I think we're quite close to releasing... To do
- [X] Set up authors (#21)
- [X] Set up version and document it (see #30)
- [X] Decide whether to use `setuptools_scm` for version number. see #28. [yes]
- [X] Expose `liberfa` version? I.e., port https://github.com/astropy/astropy/pull/9324; see #28 [yes]
- [X] ~Document the two C routines that are not part of `liberfa` - `pav2pv` and `pv2pav` (I've tried to get them up to SOFA, and probably should send a reminder).~ Remove the two extra routines, since we should not add stuff not in SOFA, and SOFA is unlikely to need it (could not find a single place where they would be useful). see #32.
- [X] Create change-log. See #34 
- [X] Add minimal usage instructions, which make distinction between ufunc and python wrapper. see #33
- [X] Check with Debian/redhat/conda, that we're making things easy enough (cc @olebole, @sergiopasra)
- [X] Document how to build with system `liberfa` library. See #28
- [X] Set up auto-release pipeline (#8; see #27)

@avalentino, @astrofrog - am I missing anything? If so, just edit this comment and add items.

avalentino:
I don't see any other blocking issue.
Maybe it could be useful to have a beta release just to be able to make an upload on pypi.

sergiopasra:
To package pyerfa in Fedora, it would be nice to support building with an external `erfa` library, just like `astropy` does. I have tried to had the following section in setup.cfg, but it doesn't work. Options are not recognized

```
[build]
use_system_libraries=1
use_system_erfa=1
```
```
error: error in setup.cfg: command 'build' has no such option 'use_system_erfa'
```

mhvk:
Right now, it is done like in `astropy`, by setting an environment variable `PYERFA_USE_SYSTEM_LIBERFA=1`. But I like the approach of a change in `setup.cfg` - just need to find out how to access those settings in the actual build process...

sergiopasra:
Well, for me setting an environment variable it's actually better. I will try to build it

mhvk:
OK, obviously we should *document* this! added on top.

olebole:
[...]

Issue #7, opened 2020-05-06
Need to set up version


mhvk:
Likely not the only thing, but needs a bit of thought on numbering: we could simply stick to the `erfa` version, but that means that if there is a bug in the python wrapper, we would need to deviate.

@avalentino, @astrofrog, @eteq - any suggestions? I guess we could let the bug-fix part deviate from `erfa` itself.

EDIT: for the post-fix, we can use the version-getting function from https://github.com/liberfa/pyerfa/pull/28#issuecomment-626300892

astrofrog:
Could we also add a fourth version number, so say ERFA is on 3.2.1, we could then have pyerfa 3.2.1, 3.2.1.1, 3.2.1.2, etc?

mhvk:
Yes, I like that fourth-number scheme!

avalentino:
+1 for the fourth-number scheme

mhvk:
Probably should at least think through the alternatives:
1. Separate version number altogether.
2. Start with `PyERFA` number and append `liberfa` (perhaps `1.1.7.0` or `1.0-1.7.0`).
Since we use the submodule, a combination seems somewhat more logical.

avalentino:
> Probably should at least think through the alternatives:
>
> 1. Separate version number altogether.
> 2. Start with PyERFA number and append liberfa (perhaps 1.1.7.0 or 1.0-1.7.0).
>    Since we use the submodule, a combination seems somewhat more logical.

IMHO (1.) is also a good option as soon as we document which is the liberfa version we bundle in each release. (2.) honestly I prefer postfix version proposed by @astrofrog , also the `1.0-1.7.0` variant could be very tricky to manage.

mhvk:
Another thought: generally, the python wrappers will/*should* work perfectly well with any `liberfa` of a given release, independent of the bug-fix number. I.e., say, `pyerfa-1.7.0` would/should work with `liberfa-1.7.1` and vice versa. So, the scheme could also be that we keep the bugfix number for `pyerfa` separate from that of `liberfa`.

Alternatively, in the `pyerfa` version, the erfa version part could simply be the *minimal* `liberfa` bugfix version it supports. Perhaps that makes more sense, since the C code that is bundled will be that minimal version, so the wrappers will work for that and simply ignore, say, any new C functions.

avalentino:
[...]

PR #28, opened 2020-05-10
Expose version information

mhvk:
This is a port of https://github.com/astropy/astropy/pull/9324/commits/e2e285a3512942141875141b6b30bb5c4ae00002, exposing the ERFA and SOFA information. Since we now have our own package, it seemed most logical to expose these inside a new `version` module - and to have that I now use `setuptools_scm` (cc @astrofrog, just to be sure I use it correctly). That solves another thing on the list of #23...

@eteq - I only expose `erfa_version` and `sofa_version`, not the erfa major, minor and micro, since I felt those were not all that useful. I also removed the version functions from `core`, since having a new package would seem reason enough to clean that up (will have to deprecate them in `astropy/_erfa`, I guess).

Since there was a missing link in `README.rst` in building the docs, I took the opportunity to mention how to install using system libraries. Another one off the list in #23...

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,5 @@
 [metadata]
 name = pyerfa
-version=1.7.0
 author = The Astropy Developers
 author_email = astropy.team@gmail.com
 license = BSD 3-Clause License
@@ -25,6 +24,7 @@ classifiers =
 packages = find:
 requires = numpy
 zip_safe = False
+setup_requires = setuptools_scm
 install_requires = numpy>=1.16
 python_requires = >=3.6
 
--- a/setup.py
+++ b/setup.py
@@ -66,12 +66,28 @@ def get_extensions():
 
 
 VERSION_TEMPLATE = """
+'''Wrapper, ERFA and SOFA version information.'''
+
+# Set the version numbers a bit indirectly, so that Sphinx can pick up
+# up the docstrings and list the values.
+from . import ufunc
+
+
+erfa_version = ufunc.erfa_version
+'''Version of the C ERFA library that is wrapped.'''
+
+sofa_version = ufunc.sofa_version
+'''Version of the SOFA library the ERFA library is based on.'''
+
+del ufunc
+
 # Note that we need to fall back to the hard-coded version if either
 # setuptools_scm can't be imported or setuptools_scm can't determine the
 # version, so we catch the generic 'Exception'.
 try:
     from setuptools_scm import get_version
     version = get_version(root='..', relative_to=__file__)
+    '''Version of the python wrappers.'''
 except Exception:
     version = '{version}'
 else:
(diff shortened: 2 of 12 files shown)


avalentino:
> Looks good, but setuptools_scm should also be added to the pyproject.toml file 

I agree. Looking at the setuptools_scm docs it should be something like::
`"setuptools>=42", "setuptools_scm[toml]>=3.4"`

setup.cfg
```diff
@@ -25,6 +24,7 @@ classifiers =
 packages = find:
 requires = numpy
 zip_safe = False
+setup_requires = setuptools_scm
```
avalentino, line 27: > if you don't plan on running setup.py commands directly, but also no harm in leaving it there in case you do want to do e.g. python setup.py --version for instance.

I think that having the possibility to run `python setup.py --version` is very handy for developers,
so i would keep the `setup_requires` entry in in `setup.cfg` (no need to add also numpy and jinja IMHO).

setup.py
```diff
@@ -65,4 +65,36 @@ def get_extensions():
...
+# version, so we catch the generic 'Exception'.
+try:
+    from setuptools_scm import get_version
+    version = get_version(root='..', relative_to=__file__)
+    '''Version of the python wrappers.'''
+except Exception:
+    version = '{version}'
+else:
+    del get_version
+""".lstrip()
```
avalentino, line 95: If you want you could also move the `VERSION_TEMPLATE` contents to a `erfa/version.py.templ` file, bit also as it is now is OK for me.

avalentino:
If in #7 we decide to go with the versioning scheme with the postfix pyerfa version number (1.7.0, 1.7.01, ...) it could be useful to customize the setuptools_scm behaviour accordingly:

```python

def guess_next_dev(version):
    from setuptools_scm.version import guess_next_version
    if version.exact:
        return version.format_with("{tag}")
    else:
        version_string = str(version.tag)
        if 'dev' in version_string or len(version_string.split('.')) > 3:
            return guess_next_version(version.tag)
        else:
            guessed = version_string.partition("+")[0] + '.1'
        return version.format_with("{guessed}.dev{distance}", guessed=guessed)

setuptools.setup(
    use_scm_version={
        ...
        'version_scheme': guess_next_dev,},
    ...)
```

setup.py
mhvk: I just copied this method over from astropy. My own sense is to stay close to that since most of us are used to it...

avalentino: OK, makes sense

mhvk:
@avalentino, @astrofrog - OK, I updated the `pyproject.toml` and for good measure also added oldest and dev dependencies test to travis.

@avalentino - thanks also for the nice piece to get the version number! I'd like to leave that for a follow-up PR, though, just to give @eteq a chance to pipe in on the version numbering options in #7.

PR #29, opened 2020-05-10
Reproducible build

avalentino:
Absolute paths used in setup.py are written into pyerfa.egg-info/SOURCES.txt making the build not reproducible.

Also fixed a small issue in linkcheck.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,9 +7,7 @@ import setuptools
 import subprocess
 
 
-ERFAPKGDIR = os.path.abspath(os.path.dirname(__file__))
-
-ERFA_SRC = os.path.abspath(os.path.join(ERFAPKGDIR, 'liberfa', 'erfa', 'src'))
+ERFA_SRC = os.path.join('liberfa', 'erfa', 'src')
 
 
 # https://mail.python.org/pipermail/distutils-sig/2007-September/008253.html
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -13,10 +13,10 @@ The project is a split of ``astropy._erfa`` module, developed in the
 context of Astropy_ project, into a standalone package.  It contains
 the ERFA_ C source code as a git submodule.
 
-.. _Python: https://www.python.org/
 .. _ERFA: https://github.com/liberfa/erfa
 .. _Numpy: https://numpy.org/
 .. _Astropy: https://www.astropy.org
+.. _Python: https://www.python.org
```
mhvk, line 19: Any particular reason to swap this around?

avalentino: Linkcheck exits with error because in `docs/quicklook.rst` there is a reference to `Python_` but the target is not included in the same file.
I disovered it making local testing.

avalentino:
> Am approving, though it would be good to remove the change in README.rst just to keep things contained.
OK, i will remove that commit. Will make a separate PR in case.

README.rst
mhvk: Ah, yes, and since you cut on astropy, now it works. OK!

PR #30, opened 2020-05-10
Set up versions with last digit a pyerfa fix number.

mhvk:
My attempt at the four-digit version number. Still not sure I actually like it all that much, since the way things are set up, the python wrappers are fairly erfa-version agnostic.

Also note the hack in the tests to work around the fact that in our self-compilation we get the wrong erfa version number (see https://github.com/liberfa/erfa/issues/62)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,18 +5,16 @@ import os
 import sys
 import setuptools
 import subprocess
+from warnings import warn
 
 
-ERFA_SRC = os.path.join('liberfa', 'erfa', 'src')
+LIBERFADIR = os.path.join('liberfa', 'erfa')
+ERFA_SRC = os.path.join(LIBERFADIR, 'src')
 
 
 # https://mail.python.org/pipermail/distutils-sig/2007-September/008253.html
 class NumpyExtension(setuptools.Extension):
     """Extension type that adds the NumPy include directory to include_dirs."""
-
-    def __init__(self, *args, **kwargs):
-        self.cython_directives = kwargs.pop('cython_directives', {})
-        super().__init__(*args, **kwargs)
 
     @property
     def include_dirs(self):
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -5,11 +5,12 @@
...
 import setuptools
 import subprocess
+from warnings import warn
 
 
 ERFAPKGDIR = os.path.abspath(os.path.dirname(__file__))
-
-ERFA_SRC = os.path.abspath(os.path.join(ERFAPKGDIR, 'liberfa', 'erfa', 'src'))
+LIBERFADIR = os.path.abspath(os.path.join(ERFAPKGDIR, 'liberfa', 'erfa'))
+ERFA_SRC = os.path.abspath(os.path.join(LIBERFADIR, 'src'))
```
avalentino, line 13: Please consider to adjust the above lines to avoid absolute paths (see also #29).

mhvk: Done.

mhvk:
I'm still not sure about the version number, but will continue that discussion in #7.

setup.py
avalentino: Please also:

```ERFA_SRC = os.path.abspath(os.path.join(LIBERFADIR, 'src'))```

should become

```ERFA_SRC = os.path.join(LIBERFADIR, 'src')```

Also the `Extension` object seems to injects paths into the egg-info/SOURCES.txt

File: setup.py [hunk 2]
```diff
@@ -5,19 +5,17 @@
...
-ERFA_SRC = os.path.join('liberfa', 'erfa', 'src')
+LIBERFADIR = os.path.join('liberfa', 'erfa')
+ERFA_SRC = os.path.join(LIBERFADIR, 'src')
 
 
 # https://mail.python.org/pipermail/distutils-sig/2007-September/008253.html
 class NumpyExtension(setuptools.Extension):
     """Extension type that adds the NumPy include directory to include_dirs."""
 
-    def __init__(self, *args, **kwargs):
```
mhvk, line 17: I removed this as we do not use cython so it seemed meaningless.

avalentino: +1

mhvk:
OK, let's get this in. With #33 I think we'll be ready for a release!

PR #31, opened 2020-05-16
Fix local docs build

avalentino:
This small patch allows to build documentation locally (without installation) :

```
$ python3 setyp.py build_ext --inplace
$ make -C docs html
```

I it also handy for debian packaging


avalentino:
Thanks @mhvk

Issue #37, opened 2020-06-02
Simplify python wrapper generation


mhvk:
(Partially inspired by an old issue I raised over at https://github.com/astropy/astropy/issues/6073)

It is not obvious one always needs all erfa functions available, and at least the python wrappers could be auto-generated (especially with the new module level `__getattr__`).

Though in fact all that do not have a warning/error flag could just omitted altogether (i.e., one would call the ufunc directly).

PR #39, opened 2020-07-06
Do not run the code generator if PYERFA_USE_SYSTEM_LIBERFA=1 (Closes #38)

### LINKED ISSUE #38: Local headers and c sources still needed despite PYERFA_USE_SYSTEM_LIBERFA
`erfa_generator.py` uses the locally provided headers and c sources despite PYERFA_USE_SYSTEM_LIBERFA is specified. The way it is described, one would expect that only the global headers form the system's `erfa-devel`, `liberfa-dev`  or whatever it is called in the distribution package would be needed. But `erfa_generator.py` is called with `ERFA_SRC` before the check for the library.

https://github.com/liberfa/pyerfa/blob/85a3794bd651e7d27d8a9fa5903858c388957a02/setup.py#L29-L40

Steps to reproduce:

```sh
rm -rf liberfa/
export PYERFA_USE_SYSTEM_LIBERFA=1
pip install .
```

Output from rpm build:
```
+ exec rpmbuild -ba --define '_srcdefattr (-,root,root)' --nosignature /home/abuild/rpmbuild/SOURCES/python-pyerfa.spec
[    3s] sh: python2: command not found
[    3s] setting SOURCE_DATE_EPOCH=1593993600
[    3s] Executing(%prep): /bin/sh -e /var/tmp/rpm-tmp.ZmaaIS
[    3s] + umask 022
[    3s] + cd /home/abuild/rpmbuild/BUILD
[    3s] + cd /home/abuild/rpmbuild/BUILD
[    3s] + rm -rf pyerfa-1.7.0
[    3s] + /usr/bin/gzip -dc /home/abuild/rpmbuild/SOURCES/pyerfa-1.7.0.tar.gz
[    3s] + /usr/bin/tar -xof -
[    3s] + STATUS=0
[    3s] + '[' 0 -ne 0 ']'
[    3s] + cd pyerfa-1.7.0
[    3s] + /usr/bin/chmod -Rf a+rX,u+w,g-w,o-w .
[    3s] + rm -rf liberfa
[    3s] + RPM_EC=0
[    3s] ++ jobs -p
[    3s] + exit 0
[    3s] Executing(%build): /bin/sh -e /var/tmp/rpm-tmp.ozRTVW
[    3s] + umask 022
[    3s] + cd /home/abuild/rpmbuild/BUILD
[    3s] + /usr/bin/rm -rf /home/abuild/rpmbuild/BUILDROOT/python-pyerfa-1.7.0-0.x86_64
[    3s] ++ dirname /home/abuild/rpmbuild/BUILDROOT/python-pyerfa-1.7.0-0.x86_64
[    3s] + /usr/bin/mkdir -p /home/abuild/rpmbuild/BUILDROOT
[    3s] + /usr/bin/mkdir /home/abuild/rpmbuild/BUILDROOT/python-pyerfa-1.7.0-0.x86_64
[    3s] + cd pyerfa-1.7.0
[    3s] + export PYERFA_USE_SYSTEM_LIBERFA=1
[    3s] + PYERFA_USE_SYSTEM_LIBERFA=1
[    3s] ++ '[' -f _current_flavor ']'
[    3s] ++ true
[    3s] + python_flavor=
[    3s] + '[' -z '' ']'
[    3s] + python_flavor=tmp
[    3s] + '[' tmp '!=' python3 ']'
[    3s] + '[' -d build ']'
[    3s] + '[' -d _build.python3 ']'
[    3s] + echo python3
[    3s] + /usr/bin/python3 setup.py build '--executable=/usr/bin/python3 -s'
[    3s] Traceback (most recent call last):
[    3s]   File "erfa_generator.py", line 682, in <module>
[    3s]     main(args.srcdir, args.output, args.ufunc, args.template_loc, args.verbose)
[    3s]   File "erfa_generator.py", line 575, in main
[    3s]     with open(erfahfn, "r") as f:
[    3s] FileNotFoundError: [Errno 2] No such file or directory: 'liberfa/erfa/erfa.h'
[    3s] Traceback (most recent call last):
[    3s]   File "setup.py", line 123, in <module>
[    3s]     ext_modules=get_extensions())
[    3s]   File "setup.py", line 31, in get_extensions
[    3s]     subprocess.run(cmd, check=True)
[    3s]   File "/usr/lib64/python3.8/subprocess.py", line 512, in run
[    3s]     raise CalledProcessError(retcode, process.args,
[    3s] subprocess.CalledProcessError: Command '['/usr/bin/python3', 'erfa_generator.py', 'liberfa/erfa/src', '--quiet']' returned non-zero exit status 1.
[    3s] error: Bad exit status from /var/tmp/rpm-tmp.ozRTVW (%build)
[    3s] 
[    3s] 
[    3s] RPM build errors:
[    3s]     Bad exit status from /var/tmp/rpm-tmp.ozRTVW (%build)
[    3s] 
[    3s] greinerT450s failed "build python-pyerfa.spec" at Mon Jul  6 09:40:03 UTC 2020.
```

avalentino:
The proposed change do not run the code generator (`erfa_generator.py`) when the user requires to use the system erfa library (`PYERFA_USE_SYSTEM_LIBERFA=1`).

NOTE: the change only works if the `core.py` and `ufunc.c` have been already generated in some way (e.g. if one uses the source tarball available on PyPi.io). It is not expected to work for a fresh git clone.
An error is raised if the `core.py` and `ufunc.c` files are not found.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,9 @@ env:
         # We include liberfa-dev so we can test using a system library,
         # and python3-astropy to run a few tests that rely on Time.
         # We need venv for a test environment and pip to install pyerfa itself.
-        - APT_DEPENDENCIES='python3-numpy python3-venv python3-pip python3-jinja2 python3-pytest-astropy liberfa-dev python3-astropy'
+        # Here, we need to include python3-attr, since the version on bionic
+        # is too old.
+        - APT_DEPENDENCIES='python3-numpy python3-venv python3-pip python3-jinja2 python3-pytest-astropy liberfa-dev python3-astropy python3-attr'
 
 jobs:
     include:
--- a/setup.py
+++ b/setup.py
@@ -6,10 +6,15 @@ import sys
 import setuptools
 import subprocess
 from warnings import warn
+from distutils.dep_util import newer
 
 
 LIBERFADIR = os.path.join('liberfa', 'erfa')
 ERFA_SRC = os.path.join(LIBERFADIR, 'src')
+GEN_FILES = [
+    os.path.join('erfa', 'core.py'),
+    os.path.join('erfa', 'ufunc.c'),
+]
 
 
 # https://mail.python.org/pipermail/distutils-sig/2007-September/008253.html
@@ -27,13 +32,26 @@ class NumpyExtension(setuptools.Extension):
 
 
 def get_extensions():
-    cmd = [sys.executable, 'erfa_generator.py', ERFA_SRC, '--quiet']
-    subprocess.run(cmd, check=True)
+    gen_files_exist = all(os.path.isfile(fn) for fn in GEN_FILES)
+    gen_files_outdated = False
+    if os.path.isdir(ERFA_SRC):
+        # assume thet 'erfaversion.c' is updated at each release at least
+        src = os.path.join(ERFA_SRC, 'erfaversion.c')
+        gen_files_outdated = any(newer(src, fn) for fn in GEN_FILES)
+    elif not gen_files_exist:
+        raise RuntimeError(
+            'Missing "liberfa" source files, unable to generate '
+            '"erfa/ufunc.c" and "erfa/core.py". '
+            'Please check your source tree. '
+            'Maybe "git submodule update" could help.')
+
+    if not gen_files_exist or gen_files_outdated:
+        print('Run "erfa_generator.py"')
+        cmd = [sys.executable, 'erfa_generator.py', ERFA_SRC, '--quiet']
+        subprocess.run(cmd, check=True)
 
     sources = [os.path.join('erfa', 'ufunc.c')]
-
     include_dirs = []
-
     libraries = []
 
     if int(os.environ.get('PYERFA_USE_SYSTEM_LIBERFA', 0)):


avalentino:
Is this also related to #37?

mhvk:
@avalentino - #37 is about the python wrappers only. But agreed with the issue raised here, and in fact a general problem: if one uses the system files, the system library may be out of date with the C code used here to generate the ufuncs. Though I guess there is at least some guarding against that by ufunc.c relying on the erfa header files (it is a bit the erfa docstrings are not part of the headers!).

About the PR proper, wouldn't it make more sense for the logic to be that `ufunc.c` and `core.py` only get generated if they do not exist already or are out of date? Though I must admit I don't know how to do the latter in an extension (would be easy in a `Makefile`...). But more simply, perhaps the logic can be that if the system library is used and the files exists, the creation step is skipped? Though then again, perhaps what you have here is cleaner.

avalentino:
> @avalentino - #37 is about the python wrappers only. But agreed with the issue raised here, and in fact a general problem: if one uses the system files, the system library may be out of date with the C code used here to generate the ufuncs. Though I guess there is at least some guarding against that by ufunc.c relying on the erfa header files (it is a bit the erfa docstrings are not part of the headers!).

if `ufuncs.c` includes new functions that are not present in the system erfa, then pyerfa should not build at all (assuming `PYERFA_USE_SYSTEM_LIBERFA=1`).
On the other hand if the system erfa is newer than the one used to generate `ufuncs.c` then in the worst case it can happen that pyerfa will miss some new erfa function (but it in any case will miss new functions if one uses bundled erfa sources to generate `ufuncs.c`).
In all cases it should never happen that you successfully build pyerfa and then it crashes at runtime (if the system liberfa does not changes).

> About the PR proper, wouldn't it make more sense for the logic to be that ufunc.c and core.py only get generated if they do not exist already or are out of date? Though I must admit I don't know how to do the latter in an extension (would be easy in a Makefile...). But more simply, perhaps the logic can be that if the system library is used and the files exists, the creation step is skipped? Though then again, perhaps what you have here is cleaner.

Just pushed a change that should go in de direction you described (I hope I have understood correctly your request).

mhvk:
@avalentino - this looks good but let me push a fix for the apt-based CI run, so that we can check it works also on our CI run. Also, this seems like a PR for which input from @sergiopasra would be useful - would be bad if it turns out that to make using the system library easier on one distribution, it becomes harder on another!

mhvk:
@avalentino - see #40 - if we put that in and you rebase, then hopefully all tests will pass here too.

avalentino:
@mhvk just rebased, now all tests seem to work.
I will wait for the merge of #40 before merging this one.

PR #40, opened 2020-07-07
Fix ci run that uses apt and system libraries

mhvk:
With this fix, the apt run should work again, and we can check that #39 works on a Debian system.


mhvk:
p.s. Had to rebase since master seems behind v1.7.0. Will try to fix that separately.

avalentino:
LGTM

Issue #35, opened 2020-05-31
Post-1.7.0 release nits


mhvk:
- [x] Authors on pypi is still "the astropy developers", as is the e-mail. Fix in #43.
- [ ] A logo would still be nice (#26)
- [x] Should also publish to Zenodo;
- [x] Add zenodo badge; see #44
- [x] Might want to have travis badge; fix in #43.
- [x]  Also a readthedocs badge? Fix in #43.
- [x] Add link to readthedocs documentation in `README.rst` (and don't expose it in docs...); fix by adding a badge in #43

mhvk:
@eteq or @astrofrog - could you enable third-party access so we can have pyerfa on zenodo? (I'm happy to do the actual linking, but it needs the access...)

astrofrog:
I wasn't sure what was meant by 3rd party access but I've now made all PyERFA maintainers also be admins for the repo instead of just maintainers.

mhvk:
Yes, it is! Now set up to automatically be stored on zenodo once another release is made. The changes above may be good enough to warrant an 1.7.0.1...

mhvk:
Actually, turns out it is triggered by publishing on github, which I hadn't done yet (just had the tag), so just published 1.7.0 to be done with it. And #44 for adding the tag, so people have a DOI to cite.

mhvk:
closing since this is done except for making a pyerfa badge, which already has its own issue

PR #43, opened 2020-07-11
Authors and badges

mhvk:
Going through the list of post-1.7 nits in #35 - counting having a read-the-docs badge as also having a link to the documentation in `README.rst`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,7 +1,6 @@
 [metadata]
 name = pyerfa
-author = The Astropy Developers
-author_email = astropy.team@gmail.com
+author = The PyERFA Developers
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -1,7 +1,6 @@
 [metadata]
 name = pyerfa
-author = The Astropy Developers
-author_email = astropy.team@gmail.com
+author = The PyERFA Developers
```
avalentino, line 3: Probably we shall also update the LICENSE.rst file accordingly

mhvk: Ah, good point! now done.

PR #45, opened 2020-07-13
Make the build reproducible

lamby:
Whilst working on the [Reproducible Builds](https://reproducible-builds.org/) effort we noticed that pyerfa could not be built reproducibly. This is due to the documentation contained the current build date.

(This was originally filed in Debian as [#964958](https://bugs.debian.org/964958).)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -26,7 +26,7 @@
 # See sphinx_astropy.conf for which values are set there.
 
 from datetime import datetime
-import sys
+import os, sys, time
 
 from pkg_resources import get_distribution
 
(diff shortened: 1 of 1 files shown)


avalentino:
Thanks @lamby

docs/conf.py
```diff
@@ -41,6 +41,9 @@
 
 project = 'PyERFA'
 author = 'The PyERFA Developers'
+build_date = datetime.utcfromtimestamp(
+    int(os.environ.get('SOURCE_DATE_EPOCH', time.time()))
+)
 copyright = '2011–{0}, {1}'.format(datetime.utcnow().year, author)
```
bmwiedemann, line 47: shouldnt this use `build_date.year` here?

avalentino: thanks @bmwiedemann you are perfectly right.
I will fix it ASAP

lamby: [The original patch from Debian is unaffected](https://bugs.debian.org/cgi-bin/bugreport.cgi?att=1;bug=964958;filename=pyerfa.diff.txt;msg=5) - it was the version here that was copied incorrectly.

PR #53, opened 2020-09-26
Liberfa config

### LINKED ISSUE #52: liberfa version number test too tight
On Debian, we have pyerfa and erfa in separate packages and update them independently (and by different people).  However, there is one test that fails when doing so: https://github.com/liberfa/pyerfa/blob/822d095c96e8203d222be23450beb83dc792b876/erfa/tests/test_erfa.py#L37-L43

Specifically, I now updated erfa to 1.7.1, and pyerfa-1.7.0 [fails on this test](https://ci.debian.net/data/autopkgtest/testing/amd64/p/pyerfa/7117273/log.gz) now. It would IMO be better if the test would just check the backward compatibility, i.e. that pyerfa uses a version that is too old (incompatible) with the current version. But in principle, if all other tests succeed (and have sufficient code coverage) there would be no need to have this test at all?

Having too strong dependencies on the package (even for testing) makes it much harder to maintain them.

Cc: @avalentino as the Debian maintainer of pyerfa

avalentino:
Improve liberfa configuration management as discussed in liberfa/erfa#73.
Also try to improve version testing to fix #52

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,11 @@
+1.7.1 (UNDER DEVELOPMENT)
+=========================
+
+- Bundled liberfa version update to v1.7.1.
+- Improved the setup machinery to ensure a proper configuration of the
+  embedded liberfa (see also https://github.com/liberfa/erfa/pull/73).
+
+
 1.7 (2020-05-31)
 ================
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,10 +1,10 @@
-1.7.1 (UNDER DEVELOPMENT)
-=========================
+1.7.1 (unreleased)
+==================
 
 - Bundled liberfa version update to v1.7.1.
 - Improved the setup machinery to ensure a proper configuration of the
-  embedded liberfa (see also https://github.com/liberfa/erfa/pull/73).
-- Improve version testing (see gh-52).
+  embedded liberfa [gh-53] (see also https://github.com/liberfa/erfa/pull/73).
+- Improve version testing [gh-53] (see issue gh-52 for rationale).
 
 
 1.7 (2020-05-31)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@
 - Bundled liberfa version update to v1.7.1.
 - Improved the setup machinery to ensure a proper configuration of the
   embedded liberfa (see also https://github.com/liberfa/erfa/pull/73).
+- Improve version testing (see gh-52).
 
 
 1.7 (2020-05-31)
--- a/erfa/tests/test_erfa.py
+++ b/erfa/tests/test_erfa.py
@@ -19,6 +19,28 @@ except ImportError:
             pass
 
 
+def embedded_liberfa(path=erfa.ufunc.__file__):
+    import platform
+    import subprocess
+
+    lddcmd = ['ldd']
+    if platform.system() == 'Darwin':
+        lddcmd = ['otool', '-L']
+    elif platform.system() == 'Windows':
+        # lddcmd = ['']  # TODO
+        return None
+
+    lddcmd.append(path)
+    try:
+        out = subprocess.run(lddcmd, check=True,
+                             encoding='utf-8', stdout=subprocess.PIPE)
+                             # capture_output=True, text=True)  # Python 3.7
+    except (subprocess.SubprocessError, OSError):
+        return None
+    else:
+        return 'erfa' not in out.stdout
+
+
 class TestVersion:
     def test_erfa_version(self):
         assert hasattr(erfa.version, 'erfa_version')
@@ -38,9 +60,15 @@ class TestVersion:
         assert hasattr(erfa, '__version__')
         version = erfa.__version__
         assert version is erfa.version.version
+
+    @pytest.mark.skipif(not embedded_liberfa(), reason='system liberfa')
+    def test_version_with_embedded_liberfa(self):
         # Oops, we had the wrong version for quite a while...
-        assert (erfa.version.erfa_version == '1.6.0' and version.startswith('1.7.0')
-                or version.startswith(erfa.version.erfa_version))
+        version = erfa.__version__
+        if erfa.version.erfa_version == '1.6.0':
+            assert version == '1.7.0'
+        else:
+            assert version.startswith(erfa.version.erfa_version)
 
 
 def test_erfa_wrapper():
--- a/erfa/tests/test_erfa.py
+++ b/erfa/tests/test_erfa.py
@@ -23,10 +23,8 @@ def embedded_liberfa(path=erfa.ufunc.__file__):
     import platform
     import subprocess
 
-    lddcmd = ['ldd']
-    if platform.system() == 'Darwin':
-        lddcmd = ['otool', '-L']
-    elif platform.system() == 'Windows':
+    lddcmd = ['nm', '--defined-only']
+    if platform.system() == 'Windows':
         # lddcmd = ['']  # TODO
         return None
 
@@ -34,11 +32,10 @@ def embedded_liberfa(path=erfa.ufunc.__file__):
     try:
         out = subprocess.run(lddcmd, check=True,
                              encoding='utf-8', stdout=subprocess.PIPE)
-                             # capture_output=True, text=True)  # Python 3.7
     except (subprocess.SubprocessError, OSError):
         return None
     else:
-        return 'erfa' not in out.stdout
+        return 'eraA2af' in out.stdout
 
 
 class TestVersion:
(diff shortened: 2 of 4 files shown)


erfa/tests/test_erfa.py
```diff
@@ -19,6 +19,28 @@ def __init__(self, *args, **kwargs):
             pass
 
 
+def embedded_liberfa(path=erfa.ufunc.__file__):
+    import platform
+    import subprocess
+
+    lddcmd = ['ldd']
```
mhvk, line 26: When I tried to check somewhat similarly on travis, I used `nm` to check the status perhaps more directly:
https://github.com/liberfa/pyerfa/blob/822d095c96e8203d222be23450beb83dc792b876/.travis.yml#L97 (here `-u` only lists undefined objects, which should be the case if the system library was used; one might want `--defined-only` for the reverse). Maybe this has an equivalent on windows too?

erfa/tests/test_erfa.py
```diff
@@ -19,6 +19,28 @@ def __init__(self, *args, **kwargs):
...
+        lddcmd = ['otool', '-L']
+    elif platform.system() == 'Windows':
+        # lddcmd = ['']  # TODO
+        return None
+
+    lddcmd.append(path)
+    try:
+        out = subprocess.run(lddcmd, check=True,
+                             encoding='utf-8', stdout=subprocess.PIPE)
+                             # capture_output=True, text=True)  # Python 3.7
```
mhvk, line 37: Is this needed, or does the above work on all supported python?

setup.py
```diff
@@ -31,11 +36,40 @@ def include_dirs(self, include_dirs):
         self._include_dirs = include_dirs
 
 
+def get_liberfa_versions(path=os.path.join(LIBERFADIR, 'configure.ac')):
+    assert os.path.exists(LIBERFADIR)
+
+    with open(path) as fd:
+        s = fd.read()
+
+    mobj = re.search(r'AC_INIT\(\[erfa\],\[(?P<version>1.7.1)\]\)', s)
```
mhvk, line 45: This regex should not have `1.7.1` in it! (rather `[0-9.]+`?)

setup.py
```diff
@@ -31,11 +36,40 @@ def include_dirs(self, include_dirs):
         self._include_dirs = include_dirs
 
 
+def get_liberfa_versions(path=os.path.join(LIBERFADIR, 'configure.ac')):
+    assert os.path.exists(LIBERFADIR)
```
mhvk, line 40: Not really important, but I'd remove this as I don't think the code can get here without it failing earlier, and even if it could, the open would fail, so there seems to be little benefit.

setup.py
```diff
@@ -65,17 +100,76 @@ def get_extensions():
...
         include_dirs=include_dirs,
         libraries=libraries,
+        define_macros=define_macros,
         language="c")
 
     return [erfa_ext]
 
 
-VERSION_TEMPLATE = """
+def _guess_next_dev(version, liberfadir=None):
```
mhvk, line 137: Over at astropy, we've changed the usage of the `setuptools_scm` quite a bit, making it simpler. What do you think? See https://github.com/astropy/astropy/pull/10774

mhvk (commented):
Finally had a look (at least at most) and it looks nice. But some in-line comments.

One general question is whether we should have the inspection of configure.ac as a fallback only, or just as the main method to get the version number. I could also see the case for just letting it depend on an environment variable - that way we can check both on travis and make sure that both continue to work.

erfa/tests/test_erfa.py
```diff
@@ -38,9 +60,15 @@ def test_version(self):
...
         version = erfa.__version__
         assert version is erfa.version.version
+
+    @pytest.mark.skipif(not embedded_liberfa(), reason='system liberfa')
+    def test_version_with_embedded_liberfa(self):
         # Oops, we had the wrong version for quite a while...
-        assert (erfa.version.erfa_version == '1.6.0' and version.startswith('1.7.0')
-                or version.startswith(erfa.version.erfa_version))
+        version = erfa.__version__
+        if erfa.version.erfa_version == '1.6.0':
```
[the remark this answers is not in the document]
avalentino: +1

erfa/tests/test_erfa.py
avalentino: nice, the command is the same both on GNU/Linux and Mac OS X, I will update.

erfa/tests/test_erfa.py
avalentino: Yes, the current implementation should work always.

setup.py
avalentino: ops, you are right, I will fix

setup.py
avalentino: +1

setup.py
avalentino: Yes, I like the solution adopted in astropy. I think it is better if I implement it in a different PR, do you agree?
In any case the machinery to guess the next dev version should be propagated in all places where ``setuptools_scm`` is used :/.

avalentino:
> One general question is whether we should have the inspection of configure.ac as a fallback only, or just as the main method to get the version number. I could also see the case for just letting it depend on an environment variable - that way we can check both on travis and make sure that both continue to work.

@mhvk honestly I would prefer to interfere as less as possible with the liberfa configuration machinery.
Inspecting ``configure.ac`` IMHO is in some way a replication of the libarfa configuration logic that should not be in pyerfa. I only added it to support (hopefully) windows users.

setup.py
mhvk: Yes, makes sense to do in a separate PR.

mhvk:
Makes sense to use `liberfa`s mechanism, though I guess taking that a bit further, one should then perhaps make the library and link to it statically? (but fine to leave that for possibly later)

On windows, should we perhaps move one of the travis runs to windows? (astropy does that, though I think with a bit of extra magic in `.travis.yml`).

avalentino:
> Makes sense to use liberfas mechanism, though I guess taking that a bit further, one should then perhaps make the library and link to it statically? (but fine to leave that for possibly later)

Indeed, it is an interesting option for future improvements. The main drawback IMHO would be the dependency on a GNU compatible make program.

> On windows, should we perhaps move one of the travis runs to windows? (astropy does that, though I think with a bit of extra magic in .travis.yml).

Yes a build on windows would surely a good thing.

I'm wondering if we should also add a test on the "sdist" target.
The logic for Liberia management has become quite complex unfortunately.

avalentino:
> I'm wondering if we should also add a test on the "sdist" target.

ah, sorry.
There is already :)

avalentino:
@mhvk if you are KO with the latest changes please go ahead and merge.

mhvk (commented):
Darn, I seem not to have pressed "submit" on my review - with just rather trivial changes.
(I can make them if you want...)

CHANGES.rst
```diff
@@ -1,3 +1,12 @@
+1.7.1 (UNDER DEVELOPMENT)
```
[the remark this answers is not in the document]
avalentino: nice!

CHANGES.rst
```diff
@@ -1,3 +1,12 @@
...
+=========================
+
+- Bundled liberfa version update to v1.7.1.
+- Improved the setup machinery to ensure a proper configuration of the
```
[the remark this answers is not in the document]
avalentino: +1

-

CHANGES.rst
```diff
@@ -1,3 +1,12 @@
...
+  embedded liberfa (see also https://github.com/liberfa/erfa/pull/73).
+- Improve version testing (see gh-52).
```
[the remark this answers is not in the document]
avalentino: +1

erfa/tests/test_erfa.py
```diff
@@ -19,6 +19,25 @@ def __init__(self, *args, **kwargs):
             pass
 
 
+def embedded_liberfa(path=erfa.ufunc.__file__):
+    import platform
+    import subprocess
+
+    lddcmd = ['nm', '--defined-only']
```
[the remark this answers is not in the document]
avalentino: yes, totally forgot to update the variable name

erfa/tests/test_erfa.py
```diff
@@ -19,6 +19,25 @@ def __init__(self, *args, **kwargs):
...
+    if platform.system() == 'Windows':
```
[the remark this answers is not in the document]
avalentino: unfortunately (or luckily :)) I don't have access to a windows machine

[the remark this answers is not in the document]
mhvk: OK, let's leave it for a todo - in principle, we can now test on travis, but I'm not sure I can be bothered...

mhvk (approved):
Nice, thanks! Merging...

PR #56, opened 2020-10-15
Test with Python 3.9

avalentino:


astrofrog:
Thanks! We might need to wait for Numpy 1.19.3 (first which will support 3.9) before all the CI passes.

avalentino:
Thanks, @astrofrog 
Do you think it makes sense to with all passes before releasing pyerfa 1.7.1?

astrofrog:
I think this could be rebased and merged now as Numpy now has a release supporting Python 3.9

avalentino:
@astrofrog the branch is already on master.
I re-triggered the test but there seems to be something strange in the test on s390x. "install with apt".


```
E: Could not configure 'libc6:s390x'. 
E: Could not perform immediate configuration on 'libcrypt1:s390x'. Please see man 5 apt.conf under APT::Immediate-Configure for details. (2)

The command "if [[ $SETUP_METHOD == 'tox' ]]; then pip install tox; else curl https://ftp-master.debian.org/keys/archive-key-10.asc | sudo apt-key add -; echo "deb http://ftp.us.debian.org/debian testing main" | sudo tee -a /etc/apt/sources.list; sudo apt-get -qq update; sudo apt-get install -y --no-install-recommends ${APT_DEPENDENCIES}; fi" failed and exited with 100 during .
```

I will in investigate.

mhvk:
@avalentino - that sounds familiar from my somewhat tortuous trials on astropy - its latest `.travis.yml` may give the clue (or I can get to it later...)

mhvk:
I think you may not need all dependencies, but good to check whether this works in any case.

Separately, we should move to `travis.com` - the `.org` will soon be decomissioned and is already superslow.

avalentino:
> I think you may not need all dependencies, but good to check whether this works in any case.

At the moment I included all the packages that were listed in APT_DEPENDENCIES.
But I agree with you that probably we could keep just some of them.
Maybe it could be enough to include:

- python3-venv  # build dependencies
- python3-pip
- liberfa-dev
- python3-numpy

> Separately, we should move to travis.com - the .org will soon be decomissioned and is already superslow.

I attempted to activate it bit probably it is necessary a liberfa maintainer to approve the request.

mhvk:
@avalentino - good, looks like test passed! 

As for the `APT_DEPENDENCIES`, previously the by-hand installation did not automatically pulled in recommended dependencies (for reasons I never understood), but those do get installed by listing packages as you do now, so I think your reduced list should suffice. Do you want to try that?

astrofrog:
It might also make sense to just switch to GitHub actions as just done for astropy core given that the free tier is virtually gone on Travis?

avalentino:
@astrofrog I have never used GitHub actions but I would like to learn. Since there is the guideline of astropy it would be not too complex for me to replicate.
I can work on it during the weekend.
In any case the change should not be blocking for this PR IMHO.

astrofrog:
Let me know if you have any questions about actions! Note that it won't run on PRs until there is a configuration in the repository so the easiest way to make sure it works the first time is to go to the Actions tab in your fork, or alternatively to push to a non-master branch in the main repo.

avalentino:
@astrofrog OK thanks for your hints. I will keep you updated.

avalentino:
@mhvk it should be ready now.
I have only added python3-pytest-astropy.
For some reason things seems not to work without it. Not clear why.

mhvk:
@astrofrog - it would seem a good idea to migrate to travis.com even if we later go for the actions, but I don't seem to have the required privilege. Do you?

PR #57, opened 2020-10-17
Version management rework

avalentino:
The version management has been reworked according to astropy/astropy#10774.
There are some differences due to the custom versioning schemed used by pyerfa, but the basic concept should be the same.

### DIFF
--- a/erfa/_dev/scm_version.py
+++ b/erfa/_dev/scm_version.py
@@ -1,0 +1,48 @@
+# Try to use setuptools_scm to get the current version; this is only used
+# in development installations from the git repository.
+
+import pathlib
+import functools
+import os.path as pth
+from warnings import warn
+
+try:
+    from setuptools_scm import git, get_version as _get_version
+    from setuptools_scm.version import guess_next_version
+
+    def _guess_next_dev(version, liberfadir=None):
+        if liberfadir is None:
+            liberfadir = pathlib.Path(
+                __file__).parent.parent.parent / 'liberfa' / 'erfa'
+
+        erfa_version = git.parse(liberfadir)
+        if not erfa_version.exact:
+            warn(f'liberfa/erfa not at a tagged release, but at {erfa_version}')
+
+        erfa_tag = erfa_version.format_with("{tag}")
+        version_string = str(version.tag)
+
+        if version.exact:
+            if not version_string.startswith(erfa_tag):
+                warn(
+                    f'tag {version_string} does not start with '
+                    f'liberfa/erfa tag {erfa_tag}')
+
+            return version_string
+
+        else:
+            if erfa_tag > version_string:
+                guessed = erfa_tag
+            elif 'dev' in version_string or len(version_string.split('.')) > 3:
+                return guess_next_version(version.tag)
+            else:
+                guessed = version_string.partition("+")[0] + '.1'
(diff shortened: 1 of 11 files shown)


erfa/_dev/scm_version.py
```diff
@@ -0,0 +1,48 @@
...
+        if not erfa_version.exact:
+            warn(f'liberfa/erfa not at a tagged release, but at {erfa_version}')
+
+        erfa_tag = erfa_version.format_with("{tag}")
+        version_string = str(version.tag)
+
+        if version.exact:
+            if not version_string.startswith(erfa_tag):
+                warn(
+                    f'tag {version_string} does not start with '
```
mhvk, line 28: Nit: put on the same line as `warn(`

erfa/_dev/scm_version.py
```diff
@@ -0,0 +1,48 @@
...
+            else:
+                guessed = version_string.partition("+")[0] + '.1'
+            return version.format_with("{guessed}.dev{distance}",
+                                       guessed=guessed)
+
+    get_version = functools.partial(_get_version,
+                                    root=pth.join('..', '..'),
+                                    version_scheme=_guess_next_dev,
+                                    relative_to=__file__)
+except Exception:
```
mhvk, line 47: I realize this comes straight from astropy, but I wonder if it isn't better to give at least a clue about what went wrong, by doing `except Exception as exc` and then `raise ImportError(...) from exc`

setup.py
```diff
@@ -140,85 +138,27 @@ def get_extensions():
...
+except FileNotFoundError:
+    guess_next_dev = None
 else:
-    del get_version, _guess_next_dev
-""".lstrip()
-
-
-guess_next_dev = functools.partial(_guess_next_dev, liberfadir=LIBERFADIR)
-
+    import types
```
mhvk, line 147: Is it still unavoidable to read and execute the file? If our tags are set up right, shouldn't it just work like astropy?

mhvk (commented):
Some total nitpicks, but mostly a larger question, whether it is really unavoidable to read the `_dev/scm_version` file - it seems to slightly beat the point...

mhvk:
p.s. Good to change all those remaining astropy references!

erfa/_dev/scm_version.py
avalentino: +1

erfa/_dev/scm_version.py
avalentino: yes, I perfectly agree

setup.py
avalentino: Well, if I make a source distribution in my working branch I get:  ``pyerfa-1.7.1.1.dev15+g3333cb6.tar.gz``.
If I remove that part that loads ``get_next_dev`` I get ``pyerfa-1.7.2.dev16+g2f17d0f.tar.gz`` instead.

avalentino:
> Some total nitpicks, but mostly a larger question, whether it is really unavoidable to read the _dev/scm_version file - it seems to slightly beat the point...

IMHO if we want to keep out versioning schema then we need it. I mean I can't figure out a smarted way to achieve the same result.
I agree with you that it intoduces some complication ...

avalentino:
OK, merging. We can eventually address remaining open points in a new PR maybe.

PR #55, opened 2020-10-11
Test pyerfa on windows as well.

mhvk:
Use the oldest deps so we do some double duty.

PR #58, opened 2020-11-14
GitHub actions

avalentino:
With this PR most of the CI jobs previously run on Travis-CI are now moved to GitHub Actions.
The only job still remaining active on Travis is the one on s390x: the architecture seem to be not available in GitHub Actions AFAIK.

The setup of GitHub Actions is based on the one of astropy and replicates quite closely the jobs previously performed on Travis-CI. 

Jobs for Mac OS X and Windows has been added (this PR basically superseded PR #55).

Edit:
GitHub actions seems to be not triggered on PRs. You can give a look to results [here](https://github.com/avalentino/pyerfa/actions/runs/363098156)

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -18,12 +18,6 @@ jobs:
 
         include:
 
-          # - name: Tests that all the basics are covered
-          #   os: ubuntu-latest
-          #   python: 3.8
-          #   toxenv: test
-          #   toxargs: -v
-
           - name: Test with oldest supported versions of our dependencies
             os: ubuntu-16.04
             python: 3.6
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -1,0 +1,106 @@
+name: CI
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  tests:
+    name: ${{ matrix.name }} [ ${{ matrix.os }} ]
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: true
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python: [3.8]
+        toxenv: [test]
+        toxargs: [-v]
+
+        include:
+
+          # - name: Tests that all the basics are covered
+          #   os: ubuntu-latest
+          #   python: 3.8
+          #   toxenv: test
+          #   toxargs: -v
+
+          - name: Test with oldest supported versions of our dependencies
+            os: ubuntu-16.04
+            python: 3.6
+            toxenv: test-oldestdeps
+            toxargs: -v
+
+          - name: Test with development versions of our dependencies
+            os: ubuntu-20.04  # currently newer than ubuntu-latest
+            python: 3.9
+            toxenv: test-devdeps
+            toxargs: -v
+
+          - name: Code style checks
+            os: ubuntu-latest
+            python: 3.x
+            toxenv: codestyle
+            toxargs: -v
+
+          - name: Documentation build
+            os: ubuntu-latest
+            python: 3.x
+            toxenv: build_docs
+            toxargs: -v
+            apt_packages: graphviz
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+        submodules: true
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python }}
+    - name: Install APT packages
+      if: matrix.apt_packages
+      run: sudo apt-get install ${{ matrix.apt_packages }}
+    - name: Install Python dependencies
+      run: python -m pip install --upgrade tox
+    - name: Run tests
+      run: tox ${{ matrix.toxargs }} -e ${{ matrix.toxenv }} -- ${{ matrix.toxposargs }}
+
+  tests_external_liberfa:
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+    env:
+      PYERFA_USE_SYSTEM_LIBERFA: 1
+    strategy:
+      fail-fast: true
+      matrix:
+        include:
+
+          - name: Tests with external liberfa
+            os: ubuntu-20.04
+            python: 3.8
+            toxenv: test
+            toxargs: -v
+            apt_packages: python3-venv python3-pip liberfa-dev python3-numpy
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+        submodules: true
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python }}
+    - name: Install APT packages
+      if: matrix.apt_packages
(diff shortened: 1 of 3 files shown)


astrofrog:
I can open a follow up to add s390x and other architectures, I think I found a way to do it.

avalentino:
> I can open a follow up to add s390x and other architectures, I think I found a way to do it.

Very nice! Please do it.

setup.py
```diff
@@ -86,6 +86,7 @@ def get_extensions():
     define_macros = []
 
     if int(os.environ.get('PYERFA_USE_SYSTEM_LIBERFA', 0)):
+        print('Using system liberfa')
```
astrofrog, line 89: Was this just for debugging or deliberately here?

astrofrog (approved):
This looks good to me! I'll let @mhvk review too though :)

setup.py
avalentino: Yes it is deliberate.
I prefer to have a track of what happens in the install log.

.github/workflows/ci_workflows.yml
```diff
@@ -0,0 +1,106 @@
...
+  pull_request:
+
+jobs:
+  tests:
+    name: ${{ matrix.name }} [ ${{ matrix.os }} ]
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: true
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
mhvk, line 14: Since all these items are are explicitly given in each entry, do these lines do anything? If not, I'd remove them.

.github/workflows/ci_workflows.yml
```diff
@@ -0,0 +1,106 @@
...
+        python: [3.8]
+        toxenv: [test]
+        toxargs: [-v]
+
+        include:
+
+          # - name: Tests that all the basics are covered
```
mhvk, line 21: I'd remove this commented-out example - I don't think it adds anything to the real ones below.

mhvk (commented):
This looks quite clear. I've not used this yet, but I assume cancelling all jobs, etc., is as easy as it was on travis?

Also, the actions do not actually seem to have run on this PR. Does that need a separate enabling or so?

And will we indeed be able to s390 and windows on this too?

.github/workflows/ci_workflows.yml
avalentino: That line allows to run base tests (with  python: [3.8] and toxenv: [test]) on all platform:
first 3 items in https://github.com/avalentino/pyerfa/actions/runs/363098156.
Than additional configuration are "include"d into the matrix for oldestdeps, devdeps, codestyle and build_docs.

.github/workflows/ci_workflows.yml
avalentino: OK, will do it.

avalentino:
> This looks quite clear. I've not used this yet, but I assume cancelling all jobs, etc., is as easy as it was on travis?
By sure there is an easy way to re-run the job.
Honestly I have not verified if there is a stop button also

> Also, the actions do not actually seem to have run on this PR. Does that need a separate enabling or so?

AFAIK actions should be enable after the merge of this PR.

> And will we indeed be able to s390 and windows on this too?

Not sure.
Probably s390x has been included initially because it is  big-endian arch.
Regarding osx and window IMHO it is not strictly necessary but it is a good idea to have them tested (especially windows).
Also I there is PR #55 whose purpose is to add a windows build if I understand correctly.

avalentino:
>> And will we indeed be able to s390 and windows on this too?
>
> Not sure.
> Probably s390x has been included initially because it is big-endian arch.
> Regarding osx and window IMHO it is not strictly necessary but it is a good idea to have them tested (especially windows).
> Also I there is PR #55 whose purpose is to add a windows build if I understand correctly.

Thinking twice both osx and window are also tested in azure-pipelines so probably drop them form GH Actions.

astrofrog:
Here's a PR to astropy core showing how to do aarch64, ppc64le and s390x: https://github.com/astropy/astropy/pull/11045 - the relevant part is https://github.com/astropy/astropy/blob/c3eb4c5752cfc273323681c80ee13a68f4059802/.github/workflows/ci_cron_weekly.yml#L53-L87

mhvk:
As I noted over at the astropy PR, running the tests on emulated hardware seems less useful to me, while I think it is important to test big-endian. So, it seems we need to keep at least that on travis.

This does make me wonder a bit what the benefit of the Github Actions is if it means adding another CI with its own quirks (or can we get rid of Azure this way?). Is there a particular reason to move away from travis? For me, once I moved to travis.com, things were again quite fast (all my projects are relatively small and do not get many PRs, i.e., they are quite like `pyerfa`).

.github/workflows/ci_workflows.yml
mhvk: Ah, that makes sense! Thanks.

astrofrog:
The free tier of Travis is essential going away. I will email astropy dev about this later unless someone beats me to it.

.github/workflows/ci_workflows.yml
avalentino: done

avalentino:
OK so, if my understanding is correct we want to keep s390x on BE HW on travis-ci.

I'm not sure about windows and osx.
If we want to keep it in GH Action that the PR is ready to merge.
Otherwise please let me know and I will quickly update it.

mhvk (approved):
Given what I understood from @astrofrog about travis over at astropy, it sounds like we want as little as possible on travis (maybe eventually move away from it altogether).

So, for now, I think we should just go with this; I think it is very clear!

Since @astrofrog already approved, I'll go ahead and merge, and then we can switch the actions on!

avalentino:
Thanks @mhvk 

Is there something else do be done before a new release?
We have already a tag for v1.7.1 but the corresponding release on pypi.io.
The next release will be 1.7.1.1 or we plan to re-tag 1.7.1?

mhvk:
@avalentino - it seems I indeed made a 1.7.1 tag, but it did not trigger any release, and hence pypi, zenodo and github all still are at 1.7.0  (maybe because some tests failed? I vaguely recall thinking that it was fortunate nothing had happened...). So, the logical thing would seem to be to delete the tag and try again.

But let me ping @astrofrog just in case - I'm still not completely sure about the auto-release feature!

mhvk:
p.s. Whatever we do, we should ensure we add the release data to `CHANGES.rst`!

astrofrog:
Il try and take a look in a few hours! Are there failures in the Azure CI for that tag?

mhvk:
I made the on Sep 17, but cannot seem to access the Azure older logs - perhaps I need to be added to the users there?

mhvk:
If it does need some particular username, via my university I seem to be registered with microsoft as `m.vankerkwijk@utoronto.ca`

astrofrog:
@mhvk - I've invited you to Azure, but for some reason I can't browse old builds (as in September for instance). It might make sense to delete the tag from GitHub and push it up again to see if there are any issues.

astrofrog:
And just to say that I now know enough to set up the wheel machinery under GitHub actions if needed, though the configuration would be a bit longer than for Azure as there are no templates for GitHub actions.

astrofrog:
(for the Actions stuff I couldn't do it for a few days though so maybe best to keep using Azure for now)

PR #65, opened 2021-02-06
Do not package generated files

avalentino:


mhvk (approved):
How weird, I did not see a message about this. I think this would definitely help avoid version problems, thanks!!

mhvk:
Upon merging the tests failed because I think some interaction with the tag that I had pushed. But with the new tag all is OK. Thanks again!!

avalentino:
Thanks @mhvk.

Actually I have just realized that ``m4/erfa-numver.m4`` is in scm so probably it is better to leave it in also if it is not strictly necessary.
Will do another PR for that.

avalentino:
Done, see PR #67

PR #67, opened 2021-02-07
Preserve m4/erfa-numver.m4

avalentino:
Small follow on of PR #65

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -19,11 +19,11 @@ exclude liberfa/erfa/config.log
 exclude liberfa/erfa/config.status
 exclude liberfa/erfa/erfa.pc
 exclude liberfa/erfa/libtool
+exclude liberfa/erfa/m4/l*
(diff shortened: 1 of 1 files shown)


mhvk (approved):
Belatedly, this makes sense, so let's get it in.

Of course, still to do is the fact that many astropy tests fail with the new ERFA...

avalentino:
> Of course, still to do is the fact that many astropy tests fail with the new ERFA...

uh, sorry, I missed it. Do you have a link?
Is it a problem in the wrapper or a matter of compatibility with the new liberfa version?

mhvk:
I think it is actual changes in `liberfa` (in particular, polar motion); see https://github.com/astropy/astropy/issues/11305 and https://github.com/astropy/astropy/issues/11305#issuecomment-775205508 - so, probably what is needed is adjustments in the tests in astropy. 

It does mean that we should add a test with astropy in our testing framework here, so that we ensure astropy does not break.

Issue #68, opened 2021-03-09
Why is the return value of this function erfa.rz() unreasonable?


ying365:
void eraRz(double psi, double r[3][3]); When I use this function, the result i get is a matrix contains true and false.

mhvk:
Sorry for the slow reply: when I try in python, it does seem to work fine:
```
In [26]: erfa.rz(np.deg2rad(60.), np.eye(3))
Out[26]: 
array([[ 0.5      ,  0.8660254,  0.       ],
       [-0.8660254,  0.5      ,  0.       ],
       [ 0.       ,  0.       ,  1.       ]])
```

ying365:
>>> import numpy as np
>>> import erfa
>>> erfa.rz(np.deg2rad(60.),np.eye(3))
array([[False, False, False],
       [False, False, False],
       [ True,  True,  True]])

Thank you for your reply. Sorry,  as you can  see , i still can not get the same result. And i  do not know what went wrong.

mhvk:
That is super weird!  Could you let me know which versions of erfa and numpy (and maybe python?) you are using?

ying365:
Erfa is 1.7.2, numpy is 1.19.2, and python is 3.6.5.

mhvk:
Hmm, only the python version is slightly suspicious, but numpy supports it and I don't see what would be in pyerfa that wouldn't. Grasping at straws here, but I assume the two inputs `np.deg2rad(60)` and `np.eye(3)` actually look OK?

ying365:
Maybe i need to rewrite this function in python. My friends still have the same result with me and our python version is different. The inputs that you assume are ok, just an angle and a matrix.

mhvk:
Totally bizarre, still. I wish I could reproduce it, since now it is so difficult to debug. I presume `erfa.rx` and `erfa.ry` give similarly strange output?

On other options: this is of course easy enough to write yourself, but if you use astropy, you can use
```
from astropy.coordinates.matrix_utilities import rotation_matrix
from astropy import units as u

rotation_matrix(60*u.deg, 'z')

# array([[ 0.5      ,  0.8660254,  0.       ],
#        [-0.8660254,  0.5      ,  0.       ],
#        [ 0.       ,  0.       ,  1.       ]])
```

ying365:
Ok, i get it. Thank you for your help. And they have the same strange result with erfa.rz, just including false and true. How incomprehensible !

mhvk:
[...]

PR #71, opened 2021-04-24
Create tests for each function based on the built-in C tests.

mhvk:
This works for all but the zero-input ufuncs.

Note: right now quite hacky, mostly pushing to see whether it passes on other platforms.


mhvk:
I updated this PR that auto-generates a `test_ufunc.py`, which runs through all the SOFA tests and does them on the python side, thus confirming things work as expected (partially inspired by the unexpected bug reported in #68, which showed that we should test all functions on all architectures).

avalentino (approved):
LGTM

mhvk:
Thanks for the quick review!

PR #72, opened 2021-04-24
Ensure that inout arguments are properly accounted for twice

### LINKED ISSUE #68: Why is the return value of this function erfa.rz() unreasonable?

mhvk:
This is in setting the types for regular ufuncs.

Fixes #68

### DIFF
--- a/erfa/ufunc.c.templ
+++ b/erfa/ufunc.c.templ
@@ -734,7 +734,12 @@ PyMODINIT_FUNC PyInit_ufunc(void)
            as these do not get copied */ #}
     {%- for func in funcs %}
     {%- if not func.user_dtype %}
-    static char types_{{ func.pyname }}[{{ func.args_by_inout('in|inout|out|ret|stat')|count }}] = { {{ func.args_by_inout('in|inout|out|ret|stat')|map(attribute='npy_type')|join(', ') }} };
+    static char types_{{ func.pyname }}[{{ func.args_by_inout('in|inout')|count
+                                         + func.args_by_inout('inout|out|ret|stat')|count }}] = {
+    {%- for arg in func.args_by_inout('in|inout') %}{{ arg.npy_type }}, {% endfor %}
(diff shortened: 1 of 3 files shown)


erfa/ufunc.c.templ
```diff
@@ -734,7 +734,12 @@ PyMODINIT_FUNC PyInit_ufunc(void)
            as these do not get copied */ #}
     {%- for func in funcs %}
     {%- if not func.user_dtype %}
-    static char types_{{ func.pyname }}[{{ func.args_by_inout('in|inout|out|ret|stat')|count }}] = { {{ func.args_by_inout('in|inout|out|ret|stat')|map(attribute='npy_type')|join(', ') }} };
```
mhvk, line 737: This previous line didn't take into account that for the ufunc, each 'inout' element is split in an 'in' and 'out' element. As a result, for `rz`, it had 
```
static char types_rz[2] = {NPY_DOUBLE, NPY_DOUBLE};
```
i.e., it missed one.  Apparently, on MacOS only the extra one became bool (some alignment perhaps?)

erfa/ufunc.c.templ
```diff
@@ -734,7 +734,12 @@ PyMODINIT_FUNC PyInit_ufunc(void)
...
+    static char types_{{ func.pyname }}[{{ func.args_by_inout('in|inout')|count
+                                         + func.args_by_inout('inout|out|ret|stat')|count }}] = {
+    {%- for arg in func.args_by_inout('in|inout') %}{{ arg.npy_type }}, {% endfor %}
```
mhvk, line 739: I'd like to move a lot of this python-generation stuff inside `erfa_generator`... But for another PR.

avalentino (approved):
LGTM

mhvk:
Thanks for the quick review, @avalentino!

PR #73, opened 2021-04-24
Update to erfa 1.7.3

mhvk:
Hmm, I tried pushing this directly to master to release 1.7.3, but it failed on windows and macos, so I presume it will fail here too.

I sort-of feel I have seen this before, and it seems the erfa release is missing the `configure` file. If so, we may need a new release there that includes it (or can we include after the fact?). @astrofrog, @avalentino - advice much appreciated!

We certainly need to update the relevant release instructions...

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -56,6 +56,9 @@ jobs:
     - name: Install APT packages
       if: matrix.apt_packages
       run: sudo apt-get install ${{ matrix.apt_packages }}
+    - name: Install packages for macOS
+      if: matrix.os == 'macos-latest'
+      run: brew install autoconf automake
     - name: Install Python dependencies
       run: python -m pip install --upgrade tox
     - name: Run tests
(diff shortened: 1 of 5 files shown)


avalentino:
The problem seems to be 
```
Can't exec "aclocal": No such file or directory at /usr/local/Cellar/autoconf/2.71/share/autoconf/Autom4te/FileUtils.pm line 274.
```

avalentino:
There was also a problem with the strange libsofa version detection.

Please give a look to my [fork](https://github.com/avalentino/pyerfa/tree/update-to-erfa-1.7.3).
With a couple of small changes all tests seems to pass.

mhvk:
@avalentino - your changes makes sense and hence I pulled them into this PR. But what I do not understand is why the autotools thing broke for 1.7.3, but worked for 1.7.2!

Anyway, if things now work well, I'll merge this and then release 1.7.3!

avalentino:
>  But what I do not understand is why the autotools thing broke for 1.7.3, but worked for 1.7.2!

Not sure, maybe some change in the VM image for OSX

mhvk:
> Not sure, maybe some change in the VM image for OSX

Ah, yes, that could be. 

And in fact, you made the same change for `erfa` - https://github.com/liberfa/erfa/pull/78#issuecomment-774536171

So, it sounds like it is just that this particular test was not done for a while...

PR #74, opened 2021-05-10
Correct dtype for ldbody

mhvk:
I've been writing a `StructuredUnit`/`StructuredQuantity` PR for astropy that allows all erfa functions to be used with quantity, and found that the dtype for `LDBODY` is actually wrong, in that it combined position and velocity in one array (which thus cannot have different units), while it should have used the `dt_pv` dtype. This corrects that.

As I was surprised the new complete tests didn't catch that, I also looked into that: turned out to be a bit of an accidental pass. That one is now corrected too.

p.s. May make a quick bugfix release for that if I get my structured quantity PR wrapped up.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,11 +1,14 @@
-1.7.4 (unreleased)
-==================
+1.7.3.1 (unreleased)
+====================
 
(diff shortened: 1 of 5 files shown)


mhvk:
THanks!

PR #76, opened 2021-05-22
Allow structured output of the right type.

### LINKED ISSUE #70: Some overrides do not quite work right if output is given
Probably mostly because they are not used in astropy, but would be good to fix:
```
import numpy as np
from erfa import ufunc as erfa_ufunc

p = np.array([[1., 2., 3.], [4., 5., 6.]])
pv = erfa_ufunc.p2pv(p)
pv  # this is OK
# array([([1., 2., 3.], [0., 0., 0.]), ([4., 5., 6.], [0., 0., 0.])],
#       dtype={'names':['p','v'], 'formats':[('<f8', (3,)),('<f8', (3,))], 'offsets':[0,24], 'itemsize':48, 'aligned':True})
erfa_ufunc.p2pv(p, out=pv)  # Oops
# TypeError: ufunc 'p2pv' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''same_kind''
```

Similarly,
```
erfa_ufunc.zpv(out=pv)
# TypeError: ufunc 'zpv' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''same_kind''
```

mhvk:
Previously, this was not allowed as it did only a general check on VOID->VOID casting, which is not allowed.

fixes #70

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+2.0.0.1 (unreleased)
+====================
+
+- The underlying universal functions in ``erfa.ufunc`` now work with an ``out``
+  argument also if the required output is a structured array. [gh-76]
+
 2.0.0 (17/05/2021)
 ==================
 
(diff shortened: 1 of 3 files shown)


avalentino (approved):
LGTM

PR #82, opened 2021-11-07
Update the OS version for the "oldest" CI run.

avalentino:
The "ubuntu-16.04" image seems to be no longer available on GHA (https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners).
Use "ubuntu-18.04" instead.


mhvk (approved):
Ah, thanks for fixing that! Merging...

PR #88, opened 2022-08-23
Switch building of wheels to GitHub Actions

astrofrog:
This uses the OpenAstronomy workflow which are designed as a replacement of the Azure Pipelines workflow. With this we can get rid of the separate Azure configuration for wheels.

I've also added building of Mac arm64 wheels which should hopefully reduce the issues being reported on that platform.

Before merging we will need to set PYPI_TOKEN in the repo secrets, but we can already see whether things work or not.


astrofrog:
@avalentino - thanks for the review! Could you create and set a PyPI token in the repository secrets? (it should be called PYPI_TOKEN). Then feel free to merge.

avalentino:
@astrofrog I do not have access to the pyerfa project on PyPI, I thing that you are the only member.

astrofrog:
Ok I'll check this evening!

avalentino:
My username on PyPI is `antonio.valentino`. If you give me access I can doit quickly

astrofrog:
@avalentino - I've added you to PyERFA on PyPI and have also added the token here, so this should be good to go.

PR #94, opened 2022-10-24
Make use of CPython limited API

astrofrog:
This is meant to be an experiment in trying to see if we can use the [Limited API](https://docs.python.org/3.10/c-api/stable.html) - doing so would mean we would only need to build one wheel for each platform (rather than one per Python version and per platform). Note that this wouldn't require any changes to the wheel building machinery as cibuildwheel recognizes this and will only build one wheel and test it on all Python versions.

I am running into the issue that ``PyArray_malloc`` is an alias of ``PyMem_RawMalloc`` which isn't in the Limited API. I was wondering if there is a way we can allocate the arrays that would be compatible with the Limited API, for example using ``PyMem_malloc`` or ``malloc``?

@mhvk - as a Numpy expert, do you have any ideas?

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -113,7 +113,7 @@ jobs:
         python -m venv --system-site-packages tests
         source tests/bin/activate
         python -m pip install --editable .[test]
-        (nm -u erfa/ufunc.cpython-*.so | grep eraA2af) || exit 1
+        (nm -u erfa/ufunc.*.so | grep eraA2af) || exit 1
         (python -c 'import erfa' 2>&1 | grep -n 'too old') > /dev/null && (echo 'liberfa too old, skipping tests'; exit 0) || python -m pytest
 
 
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -126,37 +126,16 @@ jobs:
       test_command: pytest --pyargs erfa
       targets: |
         # Linux wheels
-        - cp39-manylinux_x86_64
-        - cp310-manylinux_x86_64
-        - cp311-manylinux_x86_64
-        - cp312-manylinux_x86_64
-        - cp39-musllinux_x86_64
-        - cp310-musllinux_x86_64
-        - cp311-musllinux_x86_64
-        - cp312-musllinux_x86_64
-        - cp39-manylinux_aarch64
-        - cp310-manylinux_aarch64
-        - cp311-manylinux_aarch64
-        - cp312-manylinux_aarch64
+        - cp3*-manylinux_x86_64
+        - cp3*-musllinux_x86_64
+        - cp3*-manylinux_aarch64
         # MacOS X wheels - we deliberately do not build universal2 wheels.
         # Note that the arm64 wheels are not actually tested so we
         # rely on local manual testing of these to make sure they are ok.
-        - cp39*macosx_x86_64
-        - cp310*macosx_x86_64
-        - cp311*macosx_x86_64
-        - cp312*macosx_x86_64
-        - cp39*macosx_arm64
-        - cp310*macosx_arm64
-        - cp311*macosx_arm64
-        - cp312*macosx_arm64
+        - cp3*macosx_x86_64
+        - cp3*macosx_arm64
         # Windows wheels
-        - cp39*win32
-        - cp310*win32
-        - cp311*win32
-        - cp312*win32
-        - cp39*win_amd64
-        - cp310*win_amd64
-        - cp311*win_amd64
-        - cp312*win_amd64
+        - cp3*win32
+        - cp3*win_amd64
     secrets:
       pypi_token: ${{ secrets.pypi_token }}
--- a/erfa/ufunc.c.templ
+++ b/erfa/ufunc.c.templ
@@ -17,6 +17,13 @@
 #include "erfa.h"
 #include "erfaextra.h"
 
+// On gcc<10 we can run into the following:
+//
+//   error: dereferencing pointer to incomplete type 'PyTypeObject'
+//
+// As mentioned in https://github.com/numpy/numpy/issues/16970,
+// the workaround is to define a fake _typeobject struct.
+
 struct _typeobject {
     int dummy;
 };
--- a/erfa/ufunc.c.templ
+++ b/erfa/ufunc.c.templ
@@ -10,7 +10,7 @@
  */
 
 #define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
-#define Py_LIMITED_API 0x030700f0
+#define Py_LIMITED_API 0x030900f0
 #include "Python.h"
 #include "numpy/arrayobject.h"
 #include "numpy/ufuncobject.h"
--- a/erfa/ufunc.c.templ
+++ b/erfa/ufunc.c.templ
@@ -17,6 +17,8 @@
 #include "erfa.h"
 #include "erfaextra.h"
 
+struct _typeobject {};
+
 #define MODULE_DOCSTRING \
     "Ufunc wrappers of the ERFA routines.\n\n" \
     "These ufuncs vectorize the ERFA functions assuming structured dtypes\n" \
--- a/erfa/ufunc.c.templ
+++ b/erfa/ufunc.c.templ
@@ -210,7 +210,7 @@ static void ufunc_loop_{{ func.pyname }}(
    {#- /* if needed, allocate memory for contiguous eraLDBODY copies */ #}
    {%- if func.user_dtype == 'dt_eraLDBODY' %}
     if (copy_b) {
-        _b = PyArray_malloc(nb * sizeof(eraLDBODY));
+        _b = malloc(nb * sizeof(eraLDBODY));
         if (_b == NULL) {
             PyErr_NoMemory();
             return;
--- a/erfa/ufunc.c.templ
+++ b/erfa/ufunc.c.templ
@@ -17,7 +17,9 @@
 #include "erfa.h"
 #include "erfaextra.h"
 
-struct _typeobject {};
+struct _typeobject {
+    int dummy;
+};
 
 #define MODULE_DOCSTRING \
     "Ufunc wrappers of the ERFA routines.\n\n" \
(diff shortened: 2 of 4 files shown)


mhvk:
@astrofrog - there definitely wasn't put much thought in this, so I think we can indeed change it. It is done inside a ufunc, which means the GIL is released, so given the comment in https://github.com/numpy/numpy/blob/8d61ebc25a117337d148f1e3d96066653bd6419a/numpy/core/include/numpy/ndarraytypes.h#L359-L368, I don't think we can use the non-raw `PyMem` function. But given that the actual memory allocation typically is very small, it should not be a problem to just use `malloc`.

setup.cfg
```diff
@@ -56,3 +56,6 @@ max-line-length = 100
 
 [pycodestyle]
 max-line-length = 100
+
+[bdist_wheel]
+py_limited_api = cp37
```
pllim, line 61: I bumped Python minversion to 3.9 in #107 , so this needs updating. FYI

astrofrog: Done

astrofrog:
Switching to using malloc() seems to fix things - after a couple of other fixes this seems to work.

Note that I've also simplified the wheel building config to not split up the different Python versions into separate builds. When using the limited API, cibuildwheel will first build e.g. the Python 3.9 wheel and then just test it on 3.10, 3.11 and 3.12 (but not build it again each time). But note that for now this probably won't work because of Numpy and Python 3.12. So we might want to wait until Numpy is released before going ahead with this, or simply skip 3.12 (which doesn't matter as much because we aren't actually building 3.12 wheels anyway with this PR).

.github/workflows/ci_workflows.yml
```diff
@@ -113,7 +113,7 @@ jobs:
         python -m venv --system-site-packages tests
         source tests/bin/activate
         python -m pip install --editable .[test]
-        (nm -u erfa/ufunc.cpython-*.so | grep eraA2af) || exit 1
+        (nm -u erfa/ufunc.*.so | grep eraA2af) || exit 1
```
astrofrog, line 116: The so files are now called ufunc.abi3.so hence the change

erfa/ufunc.c.templ
```diff
@@ -10,12 +10,17 @@
...
 #define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
+#define Py_LIMITED_API 0x030900f0
 #include "Python.h"
 #include "numpy/arrayobject.h"
 #include "numpy/ufuncobject.h"
 #include "erfa.h"
 #include "erfaextra.h"
 
+struct _typeobject {
+    int dummy;
```
astrofrog, line 21: Windows complains if there are no members in the struct - the _typeobject struct is a workaround for old versions of GCC.

.github/workflows/ci_workflows.yml
```diff
@@ -126,60 +126,16 @@ jobs:
...
-        # rely on local manual testing of these to make sure they are ok.
-        - cp312*macosx_x86_64
-        - cp312*macosx_arm64
-        # No NumPy wheel for win32
-        #- cp312*win32
-        - cp312*win_amd64
-      # Required so that cp312 can grab a pre-release numpy.
-      # Can remove when cp312 is released.
-      env: |
-        CIBW_ENVIRONMENT: PIP_PRE=1
```
pllim, line 183: I don't grok the internals of erfa, so I cannot review but I wonder why you remove the extra job for py312. I split it out on purpose because py312 needs pre-release of numpy.

astrofrog: See https://github.com/liberfa/pyerfa/pull/94#issuecomment-1710775593 - with the limited API we don't build Python 3.12 wheels anymore. We just build 3.9 wheels and test them on 3.10+. Testing them on 3.12 is disabled for now because we can't have Python 3.12 as a separate job anymore with this PR.

If you would be more comfortable waiting for a Python 3.12 release of Numpy we can do that, but it's not strictly needed - if we know via other CI config that Python 3.12 works fine then testing the wheels on 3.9, 3.10 and 3.11 is probably sufficient.

astrofrog:
For anyone coming to this now, see my latest summary of changes in https://github.com/liberfa/pyerfa/pull/94#issuecomment-1710775593

.github/workflows/ci_workflows.yml
pllim: Hmm... What about using artifact upload/download for the py312? Though maybe numpy will ship py312 wheel anyway when 1.26 is released, so maybe we can just wait it out. Thanks for the clarification!

astrofrog: Happy to just wait for 1.26 final, no point overcomplicating things.

erfa/ufunc.c.templ
mhvk: Can we insist on a newer gcc? We're at gcc-13 now... I'm not quite sure what the constraints are (nor why this was not necessary before).

erfa/ufunc.c.templ
```diff
@@ -207,7 +219,7 @@ static void ufunc_loop_{{ func.pyname }}(
    {#- /* if needed, allocate memory for contiguous eraLDBODY copies */ #}
    {%- if func.user_dtype == 'dt_eraLDBODY' %}
     if (copy_b) {
-        _b = PyArray_malloc(nb * sizeof(eraLDBODY));
+        _b = malloc(nb * sizeof(eraLDBODY));
```
mhvk, line 222: I guess I am slightly confused: `PyArray_malloc` is from numpy (https://numpy.org/devdocs/reference/c-api/array.html#memory-management), not python. Is the problem that that macro uses something that is not in the limited API? Probably worth adding a comment here.

mhvk (commented):
Looks good to me but one question and one request for a comment in the code.

I agree that we should wait until py312 compatible numpy is available.

erfa/ufunc.c.templ
astrofrog: Yeah, PyArray_malloc is an alias of PyMem_RawMalloc which is not in the Python limited API (see PR description)

erfa/ufunc.c.templ
astrofrog: Let me try and see which ones fail if I remove

astrofrog: Ok so the mussllinux image doens't have a recent compiler, as well as the ubuntu-20.04 one used for oldestdeps. This is a pretty simple workaround so I think it's worth keeping rather than try and mess around with updating compilers in CI?

mhvk: Still confused that it didn't fail before, but I agree that the work-around is simple, and you have included a comment that will allow us to remove it eventually, so that's fine.

astrofrog: This is only an issue when enabling the limited API, see https://github.com/numpy/numpy/issues/16970 for more details

avalentino (approved):
LGTM

avalentino:
Thanks @astrofrog

mhvk:
@astrofrog - I think this is OK to merge once the numpy/python 3.12 issue is sorted! Thanks!

astrofrog:
@mhvk @avalentino @pllim - I just realized there is an easy way to request pre-releases on Python 3.12 as @pllim was doing before without having to split up the build matrix so have pushed a commit here (https://github.com/liberfa/pyerfa/pull/94/commits/da319b5c4447070c3481dff487945b422e3e2aed). The wheels are now tested on Python 3.12. I believe this is therefore ready to merge assuming the CI passes.

pyproject.toml
```diff
@@ -11,3 +11,9 @@ requires = [
     "numpy>=1.26.0b1; python_version>='3.12'"
 ]
 build-backend = 'setuptools.build_meta'
+
+[tool.cibuildwheel]
+
+[[tool.cibuildwheel.overrides]]
+select = "cp312-*"
+environment = "PIP_PRE=1"
```
pllim, line 19: Does this mean all the wheels now use numpy 1.26rc or just cp312?

astrofrog: Yes on 3.12 the wheels are tested against the numpy RC. Note that we don't build any Python 3.12 wheels so the build dependencies are not important there.

astrofrog:
Ah I need to add a tweak for win32, will do shortly

pyproject.toml
pllim: 🤪

pllim: p.s. Are you going to do a similar patch for `astropy`?

pllim:
numpy is going to push out wheel for win32 "soon", if you can wait a bit.

pyproject.toml
astrofrog: Haha unfortunately not, Cython's support for the limited API is very rudimentary

pllim (commented):
numpy 1.26 stable is out with py312 wheels. I think this can be simplified now.

astrofrog:
Workaround removed - this should be ready if the CI passes

astrofrog:
@mhvk @avalentino are you happy for this to be merged now?

avalentino:
yes

pyproject.toml
```diff
@@ -11,3 +11,6 @@ requires = [
     "numpy>=1.26.0b1; python_version>='3.12'"
 ]
 build-backend = 'setuptools.build_meta'
+
+[tool.cibuildwheel]
```
mhvk, line 15: Do we still need this now that numpy 1.26 is out?

mhvk (approved):
A question whether we still need to skip python 3.12, just to be sure. If it is needed, do we need an issue to remind us to remove it?

Otherwise, a very nice simplification in our wheel building, for a single, simple change. Great! Ready to (squash&) merge!

pyproject.toml
astrofrog: Oops no will remove

pyproject.toml
pllim, line 1: I think you need a rebase to pick up my merged changes to this file.

astrofrog:
@mhvk @pllim - I think this is finally ready

mhvk (approved):
OK, let's get it in! Thanks, @astrofrog!

maxnoe:
Does switching to the limited API imply anything for performance?

astrofrog:
It shouldn't as the code is virtually unchanged (the one function that was changed was an alias to malloc anyway). But if anyone finds a performance degradation let me know!

maxnoe:
Doesn't this also limit what code Cython can produce?

pllim:
Belated LGTM. Thanks!

PR #101, opened 2023-03-19
Release/pyerfa v2.0.0.2

### LINKED ISSUE #93: Trigger a new release to build Python 3.11 wheels?
Now that Python 3.11 is out, we should make a new release of PyERFA so that the Python 3.11 wheels get built.

avalentino:
Closes #93

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -25,8 +25,8 @@ jobs:
             toxargs: -v
 
           - name: Test with development versions of our dependencies
-            os: ubuntu-20.04  # currently newer than ubuntu-latest
-            python: 3.9
+            os: ubuntu-latest
+            python: 3.11
             toxenv: test-devdeps
             toxargs: -v
 
@@ -75,7 +75,7 @@ jobs:
         include:
 
           - name: Tests with external liberfa
-            os: ubuntu-20.04
+            os: ubuntu-latest
             python: 3.8
             toxenv: test
             toxargs: -v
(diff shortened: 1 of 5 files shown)


mhvk (approved):
Looks good but shouldn't we also update `.github/workflows/ci_worflows.yml`, to include an actual test with python 3.11?

avalentino:
> Looks good but shouldn't we also update `.github/workflows/ci_worflows.yml`, to include an actual test with python 3.11?

yes, you are right.
Just pushed a change to use Python 3.11 with devdeps.

Maybe we could also use a more moderns Python version for the base configuration (instead of 3.8).

mhvk:
Thanks! I think it is fine to just update the latest version tested for now.

Aside: what goes wrong with the previous version of numpy? Is that now unsupported (doesn't matter too much, since 2.0.0.1 would still be OK, but just curious).

avalentino:
Yes, I had problems building locally with numpy 1.20, but I have not investigated the reason.
According to [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html), support for numpy 1.20 has been dropped at the end of January.
Moreover I have also checked that the same change has been implemented in astropy, so I decided to not spend too much time in investigating.

avalentino:
OK; will wait until tomorrow before merging and pushing the tag to give @astrofrog a chance to comment.

astrofrog (approved):

astrofrog:
Actually the min version of numpy in setup.cfg for install requires should match the one for oldestdeps in tox.ini. Do you want to just update the min version to 1.21?

PR #103, opened 2023-03-22
Revert mininum numpy change

mhvk:
This is to ensure astropy LTS does not have to pin. Added an extra test for numpy 1.20 since that was problematic before.

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -24,6 +24,14 @@ jobs:
             toxenv: test-oldestdeps
             toxargs: -v
 
+          - name: Test with medium old supported versions of our dependencies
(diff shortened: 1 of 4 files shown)


pllim (approved):
Thanks! FWIW, LTS needs `numpy>=1.18`.

astrofrog (approved):

mhvk:
Thanks, both! I'll merge now...

p.s. Indeed, LTS needs numpy >=1.18, but we used to have 1.17, so I thought I might as well put it back to what it was. With the next ERFA release, we can make a bigger jump.

pllim:
Thanks again for the quick fix! Related follow-up PRs:

* https://github.com/astropy/astropy/pull/14567 (LTS)
* https://github.com/astropy/astropy/pull/14568 (main and 5.2.x)

PR #107, opened 2023-08-18
Stop using oldest-supported-numpy

pllim:
Because Python 3.12 does not like it. See https://github.com/astropy/astropy/actions/runs/5898725876/job/16011937369?pr=14784 from https://github.com/astropy/astropy/pull/14784 

We have to stop testing against dev pyerfa in our devdeps job downstream until this is merged. Because providing people with Python 3.12 builds is more important for astropy.

Also see https://github.com/astropy/astropy/pull/14949

cc @mhvk @eteq


avalentino (approved):

.github/workflows/ci_workflows.yml
```diff
@@ -120,16 +124,30 @@ jobs:
       test_command: pytest --pyargs erfa
       targets: |
         # Linux wheels
-        - cp*-manylinux_i686
```
pllim, line 123: Note that I also dropped 32-bit wheels here. Numpy doesn't provide 32-bit anymore.

mhvk (approved):
thanks so much! Also for the more general clean-up... I think this all looks good.

Likely, we should follow this with a release as well, at least before astropy-6.0. But that can be separate from this.

pllim:
Note that I bumped minversions and dropped 32-bit wheels. If that is a problem, let's discuss.

mhvk:
I don't think this needs a change-log entry, since it doesn't affect any of the code. For the release proper, we can have an entry that the minimum python and numpy versions have changed (perhaps upped even a bit more given NEP 29).

pllim:
Can one of you merge if you are okay with the changes? The CI is green. Thank you for the quick approvals!

Issue #109, opened 2023-08-18
Provide nightly wheels for downstream


pllim:
Would be nice if astropy can pull your dev wheel instead of building pyerfa from source every time. You already have CI set up for it, so just a matter of uploading it somewhere.

pllim:
Theoretically available but if https://github.com/astropy/astropy/pull/15543 fails to pull it, I will bug you again.

PR #113, opened 2023-09-20
Remove custom build requirements for Python 3.12

pllim:
Just to see what happens.

Blocked by:

* #115

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -129,57 +129,34 @@ jobs:
         - cp39-manylinux_x86_64
         - cp310-manylinux_x86_64
         - cp311-manylinux_x86_64
+        - cp312-manylinux_x86_64
         - cp39-musllinux_x86_64
         - cp310-musllinux_x86_64
         - cp311-musllinux_x86_64
+        - cp312-musllinux_x86_64
         - cp39-manylinux_aarch64
         - cp310-manylinux_aarch64
         - cp311-manylinux_aarch64
+        - cp312-manylinux_aarch64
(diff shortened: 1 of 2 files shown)


pllim:
Wait, what? I didn't even touch setuptools_scm stuff, though your version logic is kinda complicated in `setup.py`; is that really necessary?

avalentino:
> Wait, what? I didn't even touch setuptools_scm stuff, though your version logic is kinda complicated in `setup.py`; is that really necessary?

It is to enforce the versioning scheme that always preserves the C ERFA version.
The implementation is indeed quite complicated but at the time I didn't manage to do better.
Any suggestion/patch is very welcome of course.

pllim:
> Any suggestion/patch is very welcome of course.

I am not well-versed in version generation. But the failure is blocking downstream testing over at `astropy` given that you do not publish a dev wheel (#109). I was just going to brute force to remove the logic and let setuptools_scm do its thing.

Either that or you will have to pin `setuptools_scm<8` (https://pypi.org/project/setuptools-scm/#history) which is also not ideal.

avalentino:
> Either that or you will have to pin `setuptools_scm<8` (https://pypi.org/project/setuptools-scm/#history) which is also not ideal.

Probably at the moment it is the fastest solution, at least for the short term.
I will try to look into the issue during the week-end but I cannot guarantee, this is a very busy period for me.

pllim:
This is ready for review. Thanks!

avalentino (approved):

avalentino:
Thanks @pllim , amazing job! :)

PR #118, opened 2023-10-13
Update liberfa to v2.0.1 and prepare for release.                                                                                            

mhvk:
This is effectively the 2.0.1 release, but running it through tests first.

Because of a bugfix of `pvstar` in SOFA, one of the astropy tests fails with it. We should check how to handle that before actually releasing this.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,11 @@
-2.0.0.4 (unreleased)
-====================
+2.0.1 (2023-11-13)
+==================
 
+- Bundled liberfa version update to v2.0.1, which is derived from SOFA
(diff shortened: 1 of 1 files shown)


mhvk:
I guess the procedure might be to merge this PR (assuming tests pass), then fix the astropy failures in its `-dev` branch, and then release?

avalentino (approved):

mhvk:
Thanks for the quick review! I'll go ahead and merge this, and then make a PR for astropy to fix the failing test.

mhvk:
See https://github.com/astropy/astropy/pull/15476 for the astropy PR. Once that is in, I think we can release `pyerfa` 2.0.1.

PR #120, opened 2023-10-18
Fix PyPy compilation

astrofrog:
When compiling with PyPy, it looks like _typeobject is already defined by PyPy itself - this causes the conda-forge build to fail since it includes PyPy builds: https://github.com/conda-forge/pyerfa-feedstock/pull/31/checks?check_run_id=17819660419

This PR does the simple fix of making sure we don't redefine _typeobject if we are using PyPy which should be harmless for CPython.

I'm also adding PyPy wheels here to see if they are fast enough to build as this would be a good regression test in itself and would also be nice to provide PyPy wheels.

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -129,13 +129,18 @@ jobs:
         - cp3*-manylinux_x86_64
         - cp3*-musllinux_x86_64
         - cp3*-manylinux_aarch64
+        - pp3*-manylinux_x86_64
+        - pp3*-manylinux_aarch64
         # MacOS X wheels - we deliberately do not build universal2 wheels.
         # Note that the arm64 wheels are not actually tested so we
         # rely on local manual testing of these to make sure they are ok.
         - cp3*macosx_x86_64
         - cp3*macosx_arm64
+        - pp3*macosx_x86_64
+        - pp3*macosx_arm64
         # Windows wheels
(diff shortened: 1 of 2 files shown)


astrofrog:
The PyPy wheel builds fail without the patch here, let's make sure they pass now.

astrofrog:
For now I've made it so that we only build wheels for PyPy versions where there are Numpy wheels as I think that is a reasonable compromise (otherwise we have to build Numpy from source on PyPy and we need to add a bit of complexity to get that to work). Then we match what Numpy does and we also ensure that PyERFA does build with PyPy without going over the top.

astrofrog:
Ok I think this is fine now and ready for review.

Once/if this is merged, I would suggest tagging a new bugfix release so that we can fix the conda build (https://github.com/conda-forge/pyerfa-feedstock/pull/31)

mhvk (approved):
I hope at some point in the not too distant future we can just require newer gcc and be done with it, but for now this is very nice!

mhvk:
p.s. I'm happy to have a pyerfa 2.0.1.1 with as main `CHANGES.rst` entry that we ensured things build for PyPy.

avalentino (approved):
LGTM

mhvk:
Great! With two approvals, I'll go ahead and merge this, and make a new release that also includes a fix for #123 (which is quite serious)

Issue #119, opened 2023-10-14
Remove constraint on numpy in pyproject.toml?


mhvk:
In #111, a constraint on numpy was added to `pyproject,toml`,
https://github.com/liberfa/pyerfa/blob/1903739c3e971938c4b4e0e4994d3e3f6039ba71/pyproject.toml#L6

This has as a consequence that `pyerfa` cannot be installed from source any more using older/newer numpy, even though the code itself allows this without problem. And that contributed to it no longer being possible to use the latest version of pyerfa with astropy LTS any more (see https://github.com/astropy/astropy/pull/15481#issuecomment-1762137298 and following).

I think this is bad and I would like to remove that constraint while still ensuring that for the wheel builds we do, it is enforced, so that those work with old and new numpy. That should not be difficult but raising this issue to ensure I understand any problems before I do so...

cc @liberfa/pyerfa-maintainers, @pllim

pllim:
I *think* removing oldest-supported-numpy is necessary for Python 3.12 support. I also *think* the real problem here is you no longer provide wheels for that are compatible with `numpy<1.25` (https://github.com/liberfa/pyerfa/pull/94). https://pypi.org/project/pyerfa/#files vs https://pypi.org/project/pyerfa/2.0.0.3/#files

As long as there are compatible wheel for astropy 5.0.x, it would not need to build from source.

mhvk:
But I thought the whole point of both numpy API and #94 was "one wheel for all"!? If the wheel doesn't work for numpy < 1.25, that is also a problem.

astrofrog:
I need to look into this in more detail but I wonder if this is actually a pip version issue - that is old versions of pip don't recognise abi3 wheels?

The version of numpy in pyproject.toml should be irrelevant for other packages as it is meant to be an isolated environment.

astrofrog:
I can try and look into this tomorrow

mhvk:
Thanks! Would be nice if it is just a pip issue (and that does seem plausible).

mhvk:
@astrofrog - it is great that the problems are solved on the astropy side. I'm still a bit confused why exactly the numpy constraint for building is needed, though. It seems particular to wanting a wheel suitable for many python/numpy versions, but why should that be a general constraint on anyone installing from source rather than something we impose in our wheel-building action?

astrofrog:
[...]

PR #121, opened 2023-10-18
Unpin numpy<2 in build requirements

pllim:
I need this to test https://github.com/astropy/astropy/pull/15495

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,6 +3,6 @@ requires = [
     "setuptools",
     "setuptools_scm>=6.2",
     "jinja2>=2.10.3",
-    "numpy>=1.25,<2"
+    "numpy>=1.25"
 ]
 build-backend = 'setuptools.build_meta'


mhvk:
@astrofrog, @saimn - see also #119 - is there any actual harm in just removing the numpy constraint altogether? As is, it seems to cause pain without a clear benefit.

pllim:
Me neither... And tox does not really tell us in the log. @astrofrog said he has a cunning plan though. 🤞

mhvk:
Those constraints are (very) reasonable for building the wheels, but that does not mean they should be used in `build_requirements` generally - I think that should just exclude things that we *know* fail (which currently would seem to imply no constraint!).

So, it seems to me it is (right now), completely safe to remove the requirement as long as we can ensure that the `build_and_publish` part of `ci_workflows.yml` uses latest numpy. I guess one way of finding out is just to try: #122.

astrofrog:
See https://github.com/astropy/astropy/pull/15506 for a possible way of testing the build with dev version of dependencies

pllim:
astropy no longer needs this after https://github.com/astropy/astropy/pull/15524

PR #128, opened 2023-10-27
Include --pre flag in pip command to install developer version of numpy in wheel builds

astrofrog:
I made the same mistake over at astropy... I've included a temporary commit for now to check that numpy dev is indeed being installed


astrofrog:
Looks like this works, I'll remove the temporary commit

astrofrog:
As this is a minor fix, I'll go ahead and merge so I can test once more if the wheels now work correctly

PR #134, opened 2024-03-03
Numpy 2 compat: replace uses of NPY_MAXARGS with a compile time constant

### LINKED ISSUE #133: Nightly wheels are out of date ?
Astropy's devdeps tests are currently crashing at startup

For instance, see https://github.com/astropy/astropy/actions/runs/8119881813/job/22196363368?pr=16087

```python-traceback
...
packages/erfa/__init__.py", line 5, in <module>
    from .version import version as __version__  # noqa
  File "/home/runner/work/astropy/astropy/.tox/py312-test-devdeps/lib/python3.12/site-packages/erfa/version.py", line 25, in <module>
    from . import ufunc
RuntimeError: module compiled against ABI version 0x1000009 but this version of numpy is 0x2000000
```

From what I can see, my guess is that the latest nightlies available at https://pypi.anaconda.org/liberfa/simple are out of date and not compatible with the numpy's latest  nightly builds. To be more exact, at the time of writing, numpy nightlies were last uploaded 17 hours ago (see https://github.com/numpy/numpy/actions/workflows/wheels.yml), but pyerfa's last successful upload is from 29 hours ago (see https://github.com/liberfa/pyerfa/actions).
This would normally not be a problem since new nightlies are supposed to be uploaded well, nightly. However it looks like the automation is currently flaky and failing more often than not (see https://github.com/liberfa/pyerfa/actions). Specifically, Windows builds are failing, and it seems that nothing gets uploaded unless every build succeeds. I'd suggest changing this last rule as a quick way to mitigate the problem (I suspect most nightly users only care about Linux wheels anyway). I'm happy to open a PR on Monday if that is still relevant (there are 2 other scheduled uploads between now and then, a single succesful one would resolve the issue downstream, as far as I can tell).

I have not researched the reason why windows builds are flaky (or non deterministic, for what it's worth), but I'm happy to give a hand in solving the root problem if I can. Let me know !

neutrinoceros:
Fix #133
This is simple but fragile (it will break at runtime if numpy ever increases `NPY_MAXARGS` beyond the current value).
I'll open an alternative PR with a more complex and robust patch.

### DIFF
--- a/erfa/ufunc.c.templ
+++ b/erfa/ufunc.c.templ
@@ -572,7 +572,9 @@ static int ErfaUFuncTypeResolver(PyUFuncObject *ufunc,
 {
     int *types;
     PyArray_Descr **dtypes;
-    int types_array[NPY_MAXARGS];
+
+    /* This array needs to be at least as long as nint+nout. In practice, 32 suffices */
+    int types_array[32];
 
     if (ufunc->userloops) {
         Py_ssize_t unused_pos = 0;
@@ -752,7 +754,10 @@ PyMODINIT_FUNC PyInit_ufunc(void)
     PyArray_Descr *dt_ymdf = NULL, *dt_hmsf = NULL, *dt_dmsf = NULL;
     PyArray_Descr *dt_sign = NULL, *dt_type = NULL;
     PyArray_Descr *dt_eraASTROM = NULL, *dt_eraLDBODY = NULL;
-    PyArray_Descr *dtypes[NPY_MAXARGS];
+
+    /* This array needs to be at least as long as nint+nout. In practice, 32 suffices */
+    PyArray_Descr *dtypes[32];
+
     /* ufuncs and their definitions */
     int status;
     PyUFuncObject *ufunc;


neutrinoceros:
Actually my alternative patch using `malloc/free`, is broken.

This part works fine
```patch
diff --git a/erfa/ufunc.c.templ b/erfa/ufunc.c.templ
index a904622..85b2f9a 100644
--- a/erfa/ufunc.c.templ
+++ b/erfa/ufunc.c.templ
@@ -572,7 +572,6 @@ static int ErfaUFuncTypeResolver(PyUFuncObject *ufunc,
 {
     int *types;
     PyArray_Descr **dtypes;
-    int types_array[NPY_MAXARGS];
 
     if (ufunc->userloops) {
         Py_ssize_t unused_pos = 0;
@@ -599,10 +598,12 @@ static int ErfaUFuncTypeResolver(PyUFuncObject *ufunc,
             goto fail;
         }
         /* Copy the types into an int array for matching */
+        int* types_array = malloc(NPY_MAXARGS*sizeof(int));
         for (j = 0; j < ufunc->nargs; ++j) {
             types_array[j] = ufunc->types[j];
         }
         types = types_array;
+        free(types_array);
         dtypes = NULL;
     }
```

But this doesn't
```patch
diff --git a/erfa/ufunc.c.templ b/erfa/ufunc.c.templ
index 1bdbd17..6ef2128 100644
--- a/erfa/ufunc.c.templ
+++ b/erfa/ufunc.c.templ
@@ -753,7 +753,8 @@ PyMODINIT_FUNC PyInit_ufunc(void)
     PyArray_Descr *dt_ymdf = NULL, *dt_hmsf = NULL, *dt_dmsf = NULL;
     PyArray_Descr *dt_sign = NULL, *dt_type = NULL;
     PyArray_Descr *dt_eraASTROM = NULL, *dt_eraLDBODY = NULL;
-    PyArray_Descr *dtypes[NPY_MAXARGS];
+    PyArray_Descr *dtypes = malloc(NPY_MAXARGS*sizeof(PyArray_Descr));
     /* ufuncs and their definitions */
     int status;
     PyUFuncObject *ufunc;
@@ -940,6 +941,7 @@ PyMODINIT_FUNC PyInit_ufunc(void)
     status = PyUFunc_RegisterLoopForDescr(
         ufunc, {{ func.user_dtype }},
         ufunc_loop_{{ func.pyname }}, dtypes, NULL);
+    free(dtypes);
     if(status != 0){
         Py_DECREF(ufunc);
         goto fail;
```

(compilation fails with `error C2440: '=': cannot convert from 'PyArray_Descr *' to 'PyArray_Descr'`)
I'll admit I'm a bit out of my depth here, and generally uncomfortable with using `malloc/free` to patch anything, so I'm going to open the present patch for review and let reviewers decide if a more future-proofed solution is actually needed.

neutrinoceros:
Also: failures are thematically related in the sense that something appears to be broken in devdeps testing, but I don't think they're *caused* by my patch, are they ?

mhvk:
Hmm, the maximum number of arguments the function can get is going to be a number that depends on the erfa routines, not on numpy, so we could just determine the maximum `nin+nout` for any erfa function, or play it safe and just replace NPY_MAX_ARGS with 32, since that obviously was large enough.

neutrinoceros:
Oh I didn't dig enough to realise that the need didn't actually depend on numpy. I think hard-coding 32 here is sufficient. Let me push the simplest of patches !

mhvk:
@neutrinoceros - I think part of the problem is an incorrect numpy version constraint. I just merged #122, which gets rid of that. Can you rebase to see if that resolves the errors here?

PR #135, opened 2024-03-03
BUG: fix compatibility with numpy 2 for optional array copy (`np.array(..., copy=False)`)

neutrinoceros:
For context see
- https://github.com/astropy/astropy/pull/16142
- https://github.com/numpy/numpy/pull/25168


mhvk:
The problem is that we are not compiling right with numpy 2.0 - I included a fix for this in #137.

PR #136, opened 2024-03-03
Numpy 2 compat: replace uses of NPY_MAXARGS with a compile time constant

neutrinoceros:
Based off #134, trying to make devdeps tests green again.


neutrinoceros:
This doesn't work. I'll iterate on my fork instead of here to avoid requiring maintainer approval for each push.

PR #137, opened 2024-03-03
Numpy 2.0 compat

### LINKED ISSUE #133: Nightly wheels are out of date ?

mhvk:
This combines a fix to `tox.ini` to allow dev-deps to work (same as is used by astropy) with the fixes from #134 and #135 (former by just including the commit). cc @neutrinoceros

fixes #133
fixes #134
fixes #135

### DIFF
--- a/erfa/helpers.py
+++ b/erfa/helpers.py
@@ -14,6 +14,8 @@ from .core import ErfaWarning
 from .ufunc import get_leap_seconds, set_leap_seconds, dt_eraLEAPSECOND
 
 
+NUMPY_LT_2_0 = np.__version__.startswith("1.")
+
 _NotFound = object()
 
 
@@ -269,8 +271,8 @@ class leap_seconds:
         if hasattr(table, '__array__'):
             table = table.__array__()[list(dt_eraLEAPSECOND.names)]
 
-        table = np.array(table, dtype=dt_eraLEAPSECOND, copy=False,
-                         ndmin=1)
+        table = np.array(table, dtype=dt_eraLEAPSECOND, ndmin=1,
+                         copy=False if NUMPY_LT_2_0 else None)
 
         # Simple sanity checks.
         if table.ndim > 1:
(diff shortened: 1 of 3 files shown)


mhvk:
Assuming that all-green here will get our wheels back, I'm going to merge this. I do think it is all OK, but we can obviously do a follow-up PR if not.

neutrinoceros:
Thanks @mhvk !

PR #140, opened 2024-03-08
BLD: fix building against numpy dev

neutrinoceros:
xref https://github.com/numpy/numpy/pull/25943

This should fix the error seen in [lastest nightly builds](https://github.com/liberfa/pyerfa/actions/runs/8198609133/job/22422415350)

### DIFF
--- a/erfa/ufunc.c.templ
+++ b/erfa/ufunc.c.templ
@@ -17,6 +17,11 @@
 #include "erfa.h"
 #include "erfaextra.h"
 
+// Backported NumPy 2 API (can be removed if numpy 2 is required)
+#if NPY_ABI_VERSION < 0x02000000
+#define PyDataType_ELSIZE(descr) ((descr)->elsize)
+#endif
+
 // On gcc<10 we can run into the following:
 //
 //   error: dereferencing pointer to incomplete type 'PyTypeObject'
@@ -376,7 +381,7 @@ ufunc_loop_matches(PyUFuncObject *self,
          */
         if (types[i] == NPY_VOID && dtypes != NULL) {
             int op_descr_type_num = op_descr->type_num;
-            int dtype_elsize = dtypes[i]->elsize;
+            npy_intp dtype_elsize = PyDataType_ELSIZE(dtypes[i]);
             /*
              * MHvK: we do our own check on casting, since by default
              * all items can cast to structured dtypes (see gh-11114),
@@ -391,12 +396,12 @@ ufunc_loop_matches(PyUFuncObject *self,
                     return 0;
                 }
             }
-            else if (dtypes[i]->elsize == 1 || dtypes[i]->elsize == 12) {
+            else if (dtype_elsize == 1 || dtype_elsize == 12) {
                 /* string structured array; string argument is OK */
                 if (!((op_descr_type_num == NPY_STRING &&
-                       op_descr->elsize <= dtype_elsize) ||
+                       PyDataType_ELSIZE(op_descr) <= dtype_elsize) ||
                       (op_descr_type_num == NPY_UNICODE &&
-                       op_descr->elsize >> 2 <= dtype_elsize))) {
+                       PyDataType_ELSIZE(op_descr) >> 2 <= dtype_elsize))) {
                     return 0;
                 }
             }


neutrinoceros:
If I understand the upstream change correctly, there is no C API that works both in numpy 1 and numpy 2 here. In other words, my patch fixes building against numpy 2 but breaks building against numpy 1. In the long game, this isn't necessarily an issue since building against numpy 2 is supposed to be the one way to produce backward compatible artifacts, so we should soon require `numpy>=2.0` at build time anyway, but in the interval between now and the actual release (candidate), there's the question of how we should handle it in CI.

It should be possible (and relatively easy) to use pre-processor directives to preserve build time compat with numpy 1, so I'll try that.

neutrinoceros:
@seberg your input would be invaluable here 😅

seberg:
You should add:
```
// Backported NumPy 2 API (can be removed if numpy 2 is required)
#if NPY_ABI_VERSION < 0x02000000
#define PyDataType_ELSIZE(descr) ((descr)->elsize)
#endif
```
since you don't want to break compiling with NumPy 1.x, but otherwise looks good.

erfa/ufunc.c.templ
```diff
@@ -376,7 +376,7 @@ ufunc_loop_matches(PyUFuncObject *self,
          */
         if (types[i] == NPY_VOID && dtypes != NULL) {
             int op_descr_type_num = op_descr->type_num;
-            int dtype_elsize = dtypes[i]->elsize;
+            int dtype_elsize = PyDataType_ELSIZE(dtypes[i]);
```
seberg, line 379:
```suggestion
            npy_intp dtype_elsize = PyDataType_ELSIZE(dtypes[i]);
```
Probably also still wrong in NumPy :).  Not that it matters in practice, I don't think NumPy _actually_ allows that large voids yet.  Just setting the stage.

neutrinoceros: of course ! I lost this bit while cleaning the branch, thanks for catching my mistake !

neutrinoceros:
Thanks a lot @seberg !

pyproject.toml
```diff
@@ -3,6 +3,6 @@ requires = [
     "setuptools",
     "setuptools_scm>=6.2",
     "jinja2>=2.10.3",
-    "numpy"
+    "numpy>=1.25",
```
neutrinoceros, line 6: see https://github.com/liberfa/pyerfa/pull/122#discussion_r1517444449

pllim:
Thanks, all! Would be nice if you could trigger nightly build after merged via `workflow_dispatch`. 🙏 

https://github.com/liberfa/pyerfa/blob/7847a2f15bae91d0b5f9086ea61e0c7e1ce378ff/.github/workflows/publish.yml#L7

pyproject.toml
neutrinoceros: I think it's now an actual requirement since my patch is using a constant that's not defined in earlier versions.

seberg: I do think that `NPY_ABI_VERSION` was always defined (if it's not I should change the pattern).

FWIW, I never quite understood what is the advantage/disavantage, but overall it shouldn't matter much?  For wheel builds you want it, but it's not technically a strict requirement for building.  But if someone builds with `--no-build-isolation` it is ignored anyway...

neutrinoceros: Happy to remove it if that's what @mhvk prefers

mhvk: Let's keep it out - I'd like to introduce a minimum version only if it is actually needed.

mhvk (commented):
Thanks, @neutrinoceros! This looks good and happy that the solution from @seberg works! Only, let's not set a minimum version unless it causes failures otherwise.

mhvk (approved):
Super, thanks!

neutrinoceros:
@mhvk can you re-trigger the publish workflow manually so that it actually uploads and hopefully fix astropy devdeps too ?

mhvk:
> @mhvk can you re-trigger the publish workflow manually so that it actually uploads and hopefully fix astropy devdeps too ?

Not sure I know how to do that, cc @astrofrog...

(in the worst case, we make a new point release...)

pllim:
Please go to https://github.com/liberfa/pyerfa/actions/workflows/publish.yml and you should see an option (I don't see it here because I have no write access). It looks like this:

Screenshot 2024-03-08 132439

pllim:
Once you have done that, a new entry would appear here:

https://github.com/liberfa/pyerfa/actions/workflows/publish.yml?query=event%3Aworkflow_dispatch

mhvk:
OK, thanks, I've done that and something is afoot!

pllim:
Upload step is green. Thanks!

pllim:
For completeness, astropy devdeps job is green again. Thanks, all!

https://github.com/astropy/astropy/actions/runs/8205309742

mhvk:
Thanks, @neutrinoceros and @pllim! So nice if all I really have to do is to press a button (and even get explained which button to press).

PR #142, opened 2024-03-20
Simplify warning and error handling and make it safe for Masked

mhvk:
I started this to ensure that it would be possible to deal with `Masked` instances for which errors or warnings need to be reported (it already works fine if the only warnings or errors are for masked elements). But looking at the code, I couldn't resist also simplifying it a bit, e.g., by using f-strings.

Note that some changes are needed on the astropy side as well (`np.unique` needs to be supported by `Masked`) so I cannot add a test here yet (but have a PR on the astropy side ready for which tests pass with this erfa version).


avalentino (approved):
LGTM

PR #144, opened 2024-04-02
BLD: require numpy>=2.0.0rc1 at compile time

neutrinoceros:
As discussed in #143

This isn't just for the sake of following guidelines: it's actually helpful to anyone that would build pyerfa from source *and* would like to easily switch back and forth between versions of numpy, which, if we can't have pre-built wheels for macOS AMD (see https://github.com/liberfa/pyerfa/pull/143#issuecomment-2028825907), includes me 😅

### DIFF
--- a/erfa/ufunc.c.templ
+++ b/erfa/ufunc.c.templ
@@ -17,11 +17,6 @@
 #include "erfa.h"
 #include "erfaextra.h"
 
-// Backported NumPy 2 API (can be removed if numpy 2 is required)
-#if NPY_ABI_VERSION < 0x02000000
-#define PyDataType_ELSIZE(descr) ((descr)->elsize)
-#endif
-
 // On gcc<10 we can run into the following:
 //
 //   error: dereferencing pointer to incomplete type 'PyTypeObject'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,6 +3,6 @@ requires = [
     "setuptools",
     "setuptools_scm>=6.2",
     "jinja2>=2.10.3",
-    "numpy"
+    "numpy>=2.0.0rc1",
 ]
 build-backend = 'setuptools.build_meta'


setup.cfg
```diff
@@ -20,11 +20,11 @@ classifiers =
 [options]
 # We set packages to find: to automatically find all sub-packages
 packages = find:
-requires = numpy
+requires = numpy>=1.19.3  # keep in sync with NPY_TARGET_VERSION (setup.py)
 zip_safe = False
 tests_require = pytest-doctestplus
 setup_requires = setuptools_scm
-install_requires = numpy>=1.19
+install_requires = numpy>=2.0.0rc1
```
neutrinoceros, line 27: this line seems entirely redundant with `pyproject.toml`, however, it's been a while since I actually used `setup.cfg` so I'm only 99% confident. If that's desired, I'm game to pull the band aid and merge those two files together (which is essentially automated with the right tooling)

neutrinoceros:
I'm not convinced that failures are related. I'll try to refresh this later today and see if it's reproducible

setup.cfg
mhvk: Let's do that in a separate PR...

setup.py
```diff
@@ -96,7 +96,10 @@ def get_extensions():
     sources = [os.path.join('erfa', 'ufunc.c')]
     include_dirs = []
     libraries = []
-    define_macros = []
+    define_macros = [
```
mhvk, line 99: Why this change? It seems reasonable, but perhaps good to add why it is even needed?

mhvk (approved):
I think I should have gotten this PR in before trying to put out a new pyerfa version. I had not realized our wheel building was dependent on what exactly was going on - pretty bad form really.

I still find the whole `pip` description of building things excruciating, but to be honest cannot quite care enough any more to try to make sense of it. If this works to get the right wheels on pypi, then let's just roll with it.

setup.cfg
astrofrog: We don't want to pin numpy>2.0rc1 in ``install_requires`` - this is for runtime dependencies!

setup.cfg
```diff
@@ -20,11 +20,11 @@ classifiers =
 [options]
 # We set packages to find: to automatically find all sub-packages
 packages = find:
-requires = numpy
+requires = numpy>=1.19.3  # keep in sync with NPY_TARGET_VERSION (setup.py)
```
astrofrog, line 23: I don't think ``requires`` is recommended anymore

setup.cfg
astrofrog: (this is different from ``setup_requires`` which is the old way of specifying build-time dependencies)

astrofrog:
I think this PR would be fine if we remove ``requires`` and update ``install_requires`` to specify the minimum *runtime* dependency of numpy. @neutrinoceros - if you can make those updates, I'm happy to review tomorrow and see if we can cut another PyERFA release as @mhvk is away.

setup.cfg
neutrinoceros: Oh my you're right. I meant to touch the previous line and didn't realize my mistake in review... will fix !

neutrinoceros:
I just dropped the second commit as it was never actually necessary for this PR and actually broken (thanks @astrofrog for pointing it out). Now this PR should be minimal and will hopefully resolve the situation at astropy on release.
I'll set out to remove setup.cfg in another PR once we're out of trouble.

astrofrog (approved):

PR #145, opened 2024-04-05
MNT: migrate static metadata out of setup.cfg

neutrinoceros:
Follow up to #144
migration from `setup.cfg` to `pyproject.toml` was mostly automated using [`ini2toml`](https://pypi.org/project/ini2toml/)


neutrinoceros:
I missed that `setuptools_scm` was a runtime dependency in dev builds, and I need to figure out how this is supposed to translate in pyproject.toml without adding it as a mandatory runtime dep

avalentino (approved):
LGTM

pyproject.toml
```diff
@@ -1,9 +1,62 @@
...
+license-files = ["LICENSE.rst"]
+include-package-data = false
+
+[tool.setuptools.packages.find]
+exclude = ["erfa._dev"]
+namespaces = false
+
+[tool.pytest.ini_options]
+minversion = "4.6"
+testpaths = ['"erfa"', '"docs"', '"README.rst"']
```
mhvk, line 52: I don't think this needs the single+double quotes -- at least, astropy just has the double ones.

pyproject.toml
```diff
@@ -1,9 +1,62 @@
...
+doctest_plus = "enabled"
+doctest_optionflags = """
+NORMALIZE_WHITESPACE
+ELLIPSIS
+FLOAT_CMP
+IGNORE_EXCEPTION_DETAIL"""
+text_file_format = "rst"
+addopts = "--doctest-rst"
+xfail_strict = true
+norecursedirs = ["'erfa/_dev'"]
```
mhvk, line 62: astropy again has just the double quote and `[\\/]` instead of just the `/` (presumably so things work on windows?).

erfa/_dev/scm_version.py
```diff
@@ -6,43 +6,55 @@
 import os.path as pth
 from warnings import warn
 
-try:
-    from setuptools_scm import git, Configuration, get_version as _get_version
```
mhvk, line 10: I'm a bit confused why this is needed, given that astropy has the simpler direct `try/except` that we had here. What other attribute is being gotten that the `__getattr__` solves? Or is it just that timing now is different?

mhvk (commented):
A few small suggestions, which we might as well fix up...

pyproject.toml
neutrinoceros: I don't think escaping is necessary. Pytest's doc is a bit terse but it does mention here and there that path should be written in UNIX style, so I figure it should be handling portability natively.

erfa/_dev/scm_version.py
neutrinoceros: I didn't have any recollection of it, luckilly the answer to your question is in the commit message:
`RFC: refactor scm_version.py to survive pytest collection if setuptools_scm isn't installed`

pyproject.toml
mhvk: Indeed, presumably our tests show this is working on windows... Thanks!

erfa/_dev/scm_version.py
mhvk: So, isn't the answer then to ensure that pytest doesn't look at this directory? Is that what astropy does?

neutrinoceros: I don't know. I couldn't reproduce the problem I was trying to solve here so I just reverted the change. The PR is more self-contained as a result.

mhvk (approved):
Thanks! Let's get it in.

PR #150, opened 2024-06-02
Ensure graphviz is installed for readthedocs

mhvk:
Currently, our docs are missing the inheritance diagrams - see https://pyerfa.readthedocs.io/en/latest/api.html#class-inheritance-diagram

I had similar problems in my own packages and found that what was needed was to install `graphviz`, so that is added here.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -7,6 +7,8 @@ build:
   jobs:
     post_checkout:
       - git fetch --shallow-since=2023-01-01  || true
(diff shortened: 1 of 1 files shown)


avalentino (approved):

mhvk:
Thanks, the latest build shows it worked... https://pyerfa.readthedocs.io/en/latest/api.html#class-inheritance-diagram
(not that the inheritance diagram is very exciting...)

Issue #152, opened 2024-06-24
Add CPython 3.13 to CI when released (tracker issue)


neutrinoceros:
I ran some experiments on my fork to test pyerfa against CPython 3.13. I report that current wheels are already compatible with the GIL-enabled build for 3.13.0b2 (which is expected with `abi3` wheels, restricted to Python's limited API). 
**update**: #153 

However, the free-threaded build isn't compatible yet. Attempting to build `pyerfa` will explicitly fail with the following error
```
In file included from erfa/ufunc.c:14:
        /opt/_internal/cpython-3.13.0b1-nogil/include/python3.13t/Python.h:51:4: error: #error "The limited API is not currently supported in the free-threaded build"
```

Here's the upstream issue to wait on before we can move forward https://github.com/python/cpython/issues/111506

neutrinoceros:
Update:
https://github.com/python/cpython/issues/111506#issuecomment-2230972875 clarifies that  free-threading + limited-api compatibility will not land in Python 3.13. I suspect we might want to start experimenting with astropy sooner than 3.14.0b1, so I'm now looking for a way to disable limited-api at build time (it should still be on by default) so we can at least build pyerfa from source with free-threading whenever we're ready for it.

neutrinoceros:
This issue should be closed when CPython 3.13 (final) is added to regular CI.

mhvk:
I think #160 was the last bit in adding CPython 3.13 support, so closing this.

PR #153, opened 2024-06-24
TST: test nightly wheels against CPython 3.13 (GIL flavor)

neutrinoceros:
As I mentionned in #152, there are no incompatibilities with Python 3.13.0b2 at the moment, but enabling continuous testing will help consolidate the same effort in astropy.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -28,10 +28,16 @@ jobs:
       # For nightly wheels as well as when building with the 'Build all wheels' label, we disable
       # the build isolation and explicitly install the latest developer version of numpy as well as
       # the latest stable versions of all other build-time dependencies.
+      #
+      # pre-release pythons are included for testing but no wheels will be produced
+      # specifically for these targets since this package uses Py_Limited_API
+      # Do not hesitate to turn this setting off if testing fail for external reasons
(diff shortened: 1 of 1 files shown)


mhvk:
Not quite sure about this: isn't it good enough to test via astropy? (I'm wondering a bit more generally why we would need a test of this every day, it is not like there are that many changes to `pyerfa`!)

neutrinoceros:
Well, maybe this workflow is really satisfying two othogonal needs: regular testing and producing nightlies.
I agree that the latter doesn't really need to happen ~daily~ nightly, given the slow development pace: maybe we should instead push a new batch only on merge ?
Regular testing on the other hand I think is important to keep: sure, pyerfa doesn't move much, but numpy does, and when something affects astropy, it's always nice to immediately know whether this package is also affected or not. That said, maybe daily testing is a bit too much still. How about going weekly ?

This particular PR shouldn't be merged as is now: upstream changes have made it at least partially redundant. Let me draft this for now. I'll revisit soon.

neutrinoceros:
For posterity, this is done in #157

neutrinoceros:
Now that numpy has cp313 wheels, this is completely unnecessary. Closing now.

PR #155, opened 2024-08-13
BLD: fix a conversion error (char[2] -> int)

neutrinoceros:
Tentatively fix a build error seen in CI
[example logs](https://github.com/liberfa/pyerfa/actions/runs/10365163862/job/28691789638)

I couldn't reproduce this locally yet so fingers crossed.


neutrinoceros:
Ok I think I got it this time: wheel builds is green on macOS arm64.
Meanwhile, we're seeing another, unrelated error on windows: following upstream changes in cibuildwheel, our wheels are being tested on CPython 3.13 too, but it doesn't quite work since numpy doesn't have stable wheels for it yet. I'll fix this in another PR.

neutrinoceros:
ping @mhvk for review

setup.py
```diff
@@ -128,8 +128,22 @@ def get_extensions():
             if liberfa_versions:
                 print('Configure liberfa ("configure.ac" scan)')
                 lines = []
-                for name, value in get_liberfa_versions():
-                    lines.append(f'#define {name} "{value}"')
```
mhvk, line 132: I think this can be done more simply as,
```
if isinstance(value, str):
    # strings should be escaped
     value = f'"{value}"'
lines.append(f'#define {name} {value}')
```
Or, just accounting for the fact that `repr` gives the wrong type of quote,
```
lines.append(f'#define {name} {value!r}'.replace("'", '"'))
```

mhvk (commented):
Thanks for finding that! Since `get_liberfa_versions()` actually returns the right types, it may make sense to just use that information, which would be a smaller change. Suggestions inline...

setup.py
neutrinoceros: neat. I just pushed your one-liner. Thanks !

mhvk:
Great, let's get this in!

PR #156, opened 2024-08-13
WHL: skip tests on CPython 3.13 for now (wait for numpy wheels)

neutrinoceros:
Rationale:
Following upstream changes in cibuildwheel, our wheels are being tested on CPython 3.13 too, but it doesn't quite work since numpy doesn't have stable wheels for it yet.

Originally posed in #155: https://github.com/liberfa/pyerfa/pull/155#issuecomment-2285647864

[example logs](https://github.com/liberfa/pyerfa/actions/runs/10365939408/job/28694130328)

Currently based off #155

PR #154, opened 2024-07-16
BLD: allow building without Py_LIMITED_API

neutrinoceros:
This is to address https://github.com/liberfa/pyerfa/issues/152#issuecomment-2231243236
This changes nothing by default but it *allows* building pyerfa without `Py_LIMITED_API` for testing purposes. This is needed to enable testing astropy in free-threaded Python.

Note that I also ran the test suite locally with free-threading (using Python 3.13.0b3 + numpy 2.1.0dev0) and got no errors so this seems like a safe bet.

PR #157, opened 2024-08-13
WHL: upload and test 'nightlies' only once a week

neutrinoceros:
As discussed in #154 
I picked Sunday because it conveniently is ~1 day before astropy's weekly cron, so it's less likely that issues get discovered downstream instead of here.

Failures are expected until #155 and #156 are merged

cc @mhvk

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,8 +2,8 @@ name: Wheel building
 
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


mhvk:
Makes sense to me - but best indeed after doing the others!

mhvk:
Let's get this one in!

PR #158, opened 2024-08-20
BLD: automatically detect if in freethread build of Python

tacaswell:
Following suggestion from:

https://docs.python.org/3.13/howto/free-threading-extensions.html#limited-c-api-and-stable-abi

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,11 +21,9 @@ GEN_FILES = [
 ]
 
 
-# build with Py_LIMITED_API unless explicitly disabled
-USE_PY_LIMITED_API = (
-    (not sysconfig.get_config_var("Py_GIL_DISABLED")) and
-    os.getenv("PYERFA_LIMITED_API", "1") != "0"
-)
+# build with Py_LIMITED_API unless in freethreading build (which does not currently
+# support the limited API in py313t)
+USE_PY_LIMITED_API = not sysconfig.get_config_var("Py_GIL_DISABLED")
 
 class bdist_wheel_abi3(bdist_wheel):
     def get_tag(self):
--- a/setup.py
+++ b/setup.py
@@ -10,6 +10,7 @@ import subprocess
 from warnings import warn
 import packaging.version
 from wheel.bdist_wheel import bdist_wheel
+import sysconfig
 
 
 LIBERFADIR = os.path.join('liberfa', 'erfa')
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -19,8 +20,12 @@
     os.path.join('erfa', 'ufunc.c'),
 ]
 
+
 # build with Py_LIMITED_API unless explicitly disabled
-USE_PY_LIMITED_API = os.getenv("PYERFA_LIMITED_API", "1") != "0"
+USE_PY_LIMITED_API = (
+    (not sysconfig.get_config_var("Py_GIL_DISABLED")) and
```
mhvk, line 26: Thanks for the PR! The link you give makes sense, but I'm still confused - this piece is about building, not running, and might we not want to build without the limited API on a regular, GIL-enabled python? cc @neutrinoceros, since he just put this in.

neutrinoceros: I don't know of a reason to build with the GIL and without Limited API, but the reason I specifically didn't implement this automatic switch is that it's only expected to be relevant for Python 3.13t, as 3.14t is expected to support the Limited API. _That said_, I made this choice before the guide had a recommendation there (or maybe before I learned about it), and maybe it's easier to follow a convention than having our own solution in place.

tacaswell: My intent with this change was to preserve the ability to build with GIL and with not-limited.

As a random person trying to build packages (https://github.com/tacaswell/build_the_world), not having to set an pyerfa specific ENV is very convenient!

mhvk: OK, that makes sense - thanks, @neutrinoceros and @tacaswell! 

@neutrinoceros - do we still need your special variable?

neutrinoceros: It's not used anywhere yet, and @tacaswell´s argument makes a lot of sense, so I say let's roll with it. Thanks !

mhvk: @neutrinoceros - So, remove your environment variable then in favor of this? (Right now, setting the environment variable is still kept as an option.)

neutrinoceros: Yes !

mhvk: @tacaswell - are you able to make the small change requested? (Remove support for @neutrinoceros's environment variable.) If not, we can do it.

mhvk (approved):
Thanks, this is enough - the environment variable was not used anywhere in the repository yet.

mhvk:
@neutrinoceros - I had been waiting for a new SOFA release, but if there are problems that can be solved now, obviously we should go ahead, so I made a point release now. Thanks for the heads-up!

PR #162, opened 2024-10-10
BLD: prefer to use bdist_wheel from setuptools rather than wheel

tacaswell:
Due to upcoming changes to
setuptools (https://github.com/pypa/setuptools/pull/4647) using bdist_wheel from wheel will fail.

I suspect that this could be simplified by setting a high enough floor on setuptools and dropping wheel as build dependencies, but I am not sure how the window of support of bdist_wheel in setuptools maps to your window of versions of setuptools you want to support.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,8 +9,14 @@ import setuptools
 import subprocess
 from warnings import warn
 import packaging.version
-from wheel.bdist_wheel import bdist_wheel
 import sysconfig
+
+try:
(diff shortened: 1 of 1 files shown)


neutrinoceros:
I think it's fine to use a try/except here but we should probably have a comment stating which version of setuptools first supported this.

neutrinoceros (approved):

avalentino (approved):
LGTM

Issue #165, opened 2025-01-24
Move setuptools_scm info to pyproject.toml


pllim:
Not sure why it happens now but s390x job in astropy started failing and specifically complains about pyerfa pyproject.toml does not contain a tool.setuptools_scm section.

Can you please move the setuptools_scm stuff to pyproject.toml tool.setuptools_scm section and make a release, and then we can see if the job runs again (or hits a different error)?

Also see:

* https://github.com/astropy/astropy/issues/17664
* https://github.com/astropy/astropy/issues/17663

neutrinoceros:
Hum, I remember seeing this warning before, so it's not new. Additionally, I checked that neither `setuptools` nor `setuptools-scm` were updated on `s390x` between the latest successful run and the earliest failing one. So I'm not confident that addressing this will indeed fix building on this arch, but at the very least it's an annoying red herring so let's fix it anyway and hope for the best.

avalentino:
Closed via #166

PR #166, opened 2025-01-25
MNT: add static configuration for `setuptools-scm`

neutrinoceros:
Address #165

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -198,7 +198,6 @@ else:
                                            liberfadir=LIBERFADIR)
 
 use_scm_version = {
-    'write_to': os.path.join('erfa', '_version.py'),
     'version_scheme': guess_next_dev,
 }
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [build-system]
 requires = [
     "setuptools>=61.2",
-    "setuptools_scm>=6.2",
+    "setuptools_scm>=8.0.0",
     "wheel",
     "jinja2>=2.10.3",
     "numpy>=2.0.0rc1",
@@ -46,6 +46,9 @@ include-package-data = false
 [tool.setuptools.packages.find]
 exclude = ["erfa._dev"]
 namespaces = false
+
+[tool.setuptools_scm]
+version_file = "erfa/_version.py"
 
 [tool.pytest.ini_options]
 minversion = "4.6"


neutrinoceros:
good news, it looks like we can actually reproduce the real error on this repo now: https://github.com/liberfa/pyerfa/actions/runs/12963097535/job/36160295587

neutrinoceros:
The build error on aarch64 is unrelated. See upstream issue at https://github.com/tonistiigi/binfmt/issues/215

pllim:
Is it also possible to clean up setup.py version stuff? Do we really need guess next version?

neutrinoceros:
Oh I missed that part. I was assuming that `setuptools-scm` was relying on some default behavior but it actually is all done dynamically... you're right, I do need to address this in the same PR !

>  Do we really need guess next version?

Indeed we shouldn't need this. I'm guessing it was done to support building without setuptools-scm available 4 years ago (https://github.com/liberfa/pyerfa/pull/57/commits/65603bd3c8d33079caf831a69780c27bb52803a8)

pllim:
There's also a version.py but it's probably ok as is. Second pair of eyes won't hurt. Thanks!

mhvk:
This looks good but as I don't remember quite why we had the fairly complicated code in `setup.py`, I've requested a review from @avalentino.

mhvk:
p.s. Thanks for linking to the upstream issue with `aarch64`.

avalentino:
Dear @neutrinoceros , @pllim , @mhvk ,
I fear that not using a proper `guess_next_dev` function in `setup.py` would break the version scheme described in [RELEASING.rst](https://github.com/liberfa/pyerfa/blob/main/RELEASING.rst) (that is also enforced in some of the tests).
Even if no tests fails right now, I fear that we would have failures at the next update of the C `liberfa` version.

According to my understanding you want to have
```
[tool.setuptools_scm]
version_file = "erfa/_version.py"
```
in the `pyproject.toml` file.
IMHO this is still possible without dropping also 
```
use_scm_version = {
    'version_scheme': guess_next_dev,
}
```
from `setup.py`.
By the way, I would be in favour of a simplification of the current `setuptools-scm` machinery, and I would be open to discuss a change in the versioning scheme if you want to propose something,

setup.py
```diff
@@ -179,31 +178,7 @@ def get_extensions():
...
-        exec(code, scm_version.__dict__)
-    except ImportError:
-        guess_next_dev = None
-    else:
-        guess_next_dev = functools.partial(scm_version._guess_next_dev,
-                                           liberfadir=LIBERFADIR)
-
-use_scm_version = {
-    'write_to': os.path.join('erfa', '_version.py'),
-    'version_scheme': guess_next_dev,
```
avalentino, line 202: Not using `g.ess_next_dev` in `setup.py` would break the versioning scheme at the next update of the `liberfa` C library version,

avalentino (changes requested):

neutrinoceros:
Thanks @avalentino for your quick review !
I reverted most of the changes in `setup.py`, now this PR is only about moving the deprecated `write_to` option to `pyproject.toml` and replacing it with it's not-deprecated equivalent. I checked that setuptools-scm knows how to combine static and dynamic configuration options by setting `version_file = "erfa/_version_test.py"` and inspecting build artifacts.

avalentino (approved):

avalentino:
Thanks @neutrinoceros

pllim:
Thanks all!

pllim:
A release soon would be nice so the "exotic" job can pick it up. We don't run those jobs on dev pyerfa.

mhvk:
Hmm, the `aarch64` failure also causes the wheel building generally to be aborted, so I cannot really release the wheels. I could do a github source release, but am not sure that is useful...

pllim:
Hmm is it possible to opt out of aarch64 wheel for *this* release?

avalentino:
> OK, I'm making a new release. I think the release gets into `ubuntu-rolling` via Debian, i.e., really via @olebole.
> 
> p.s. I guess the wheels built will fail for `aarch64`...

I can help for the debian part. Once the release is ready I can quickly push the new version in debian

PR #168, opened 2025-05-19
WHL(EXP): test cibuildwheel 3.0.0b1

neutrinoceros:
testing out https://github.com/OpenAstronomy/github-actions-workflows/pull/278, and adjusting settings:
pypy builds are now disabled by default, and `pp39` requires `enable: pypy-eol`, which I think is a sign that we should start building against more recent versions of pypy instead, though, for the time being, I'm doing a simple dependency upgrade with no observable change.


neutrinoceros:
Looks like the wheel-building workflow was automatically disabled, which defeats the purpose of this PR.
Could @mhvk or @avalentino turn it back on ?

mhvk:
@neutrinoceros - I'm trying to parse the workflows, but I actually do not understand why the wheel-building is not run, nor know how to enable it. Let me ping @astrofrog, who I think implemented it (and who certainly understands github actions better).

neutrinoceros:
This seems to be going well, so the upgrade should be trivial when the stable release is out.
What do you think about upgrading PyPy wheels to pp310 ? this version has been available for almost 2 years and doesn't seem about to get away anytime soon. I could do this in a separate PR.

neutrinoceros:
Whoops, got a build error on windows and no time to fix it right now, sorry about the noise. I'll try to circle back to it later today !

avalentino (approved):

PR #173, opened 2025-07-30
MNT: adjust license spec to PEP 639

neutrinoceros:


avalentino:
Thanks @neutrinoceros

Issue #175, opened 2025-08-14
Suggestion: let pyerfa functions return NamedTuple whenever there is more than 1 result


mhvk:
@eerovaher suggested at https://github.com/astropy/astropy/pull/18469#discussion_r2277705254 that it would make propagating values from one erfa function to another substantially more readable if erfa functions that return more than one argument return a `NamedTuple`. This may be tricky for the `ufunc` implementations themselves, but is easy to do for all functions that are currently wrapped (those that check warning and error codes); possibly, this is a good enough reason to wrap all `ufunc`.

eerovaher:
I brought this up in astropy/astropy#18469, but the second commit in astropy/astropy#18522 is a better example of how a `NamedTuple` could be used. There the `NamedTuple` subclass is defined in `astropy`, but it is not obvious to be why it couldn't be defined in `pyerfa`.

PR #176, opened 2025-08-17
Use NamedTuple for functions with multiple outputs

mhvk:
Following up on #175, this PR lets functions with more than one output return `NamedTuple`, with the names as they are in the docstring.

Notes:
1. I did a little not-quite-related docstring and template refactoring in the first two commits; can split out if wanted.
2. This might be nicer with a new ERFA release, but that has been a while...
3. Typing is on purpose kept minimal, as I think that is better done in a separate PR. This would involve:
  - Add dtypes for the new `NamedTuple` attributes.
  - Add input and return types to all functions.
  - Possibly expose the new `NamedTuple` classes (not sure; neither astropy nor numpy does it).

p.s. I checked that sphinx renders the new `Returns` sections nicely.


mhvk:
We need to increase our minimum supported numpy if we want to type as `NDArray`.

PR #181, opened 2025-10-15
Implement `leap_seconds` as a module

eerovaher:
[PEP 562 – Module `__getattr__` and `__dir__`](https://peps.python.org/pep-0562/) allows re-implementing `leap_seconds` as a module while retaining backwards compatibility. The main advantage is that modules are naturally singletons which is important because `pyerfa` uses a single `leap_seconds` object to manage leap seconds globally.


avalentino:
Thanks @eerovaher

PR #215, opened 2026-03-30
Freethreading

ColmTalbot:
The changes I needed were

- making erfa leap seconds thread safe
- making the leap second expiration global variables thread local
- I tested the impure `aper` functions, but those seem to not actually modify the input in place in pyerfa, so those are safe as far as I can tell.

I ran all of the tests with [pytest-run-parallel](https://github.com/quansight-labs/pytest-run-parallel) and they pass (except the docstring test, which pytest-run-parallel can't handle).

Remaining questions I have:

- should I write something up in the docs somewhere?
- I added a py314t test to the CI suite, is that alright? Should I add more tests?

See https://github.com/liberfa/pyerfa/issues/209 cc @neutrinoceros @mhvk 

Needs https://github.com/liberfa/erfa/pull/108.

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -52,7 +52,7 @@ jobs:
           - name: Test with freethreaded python
             os: ubuntu-latest
             python: '3.14t'
-            toxenv: test-devdeps
+            toxenv: test-freethreaded
             toxargs: -v
 
           - name: Documentation build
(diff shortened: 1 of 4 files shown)


avalentino:
liberfa/erfa#108 has been merged.
We are now waiting for a new liberfa release to be able to merge this one.

mhvk:
Yes, absolutely, and it just means another far down the line point release. Main problem on my side that is that teaching has started and I'm rather short on time. But happy to approve/merge if someone else takes the lead...

PR #220, opened 2026-04-03
Rewrite version tests

eerovaher:


erfa/tests/test_erfa.py
```diff
@@ -34,31 +34,29 @@ def check_embedded_liberfa():
     return None if b"eraA2af" in out else pytest.skip(reason="system liberfa")
 
 
-class TestVersion:
```
eerovaher, line 37: The tests in this class do not share any setup or teardown, so the only reason for bundling them together in a class that I can think of is to allow running only version tests with `pytest erfa/tests/test_erfa.py::TestVersion`. But in practice `pytest -k version` works just as well and does not require implementing a separate class. If it really is important to separate version tests from others then they should be in a separate module.

erfa/tests/test_erfa.py
```diff
@@ -34,31 +34,29 @@ def check_embedded_liberfa():
...
-    def test_erfa_version(self):
-        assert hasattr(erfa.version, 'erfa_version')
```
eerovaher, line 39: These `assert hasattr()` checks do not serve a useful purpose. If the attribute is missing then trying to access it will cause an `AttributeError`, but `pytest` can catch and report those just fine.

erfa/tests/test_erfa.py
```diff
@@ -34,31 +34,29 @@ def check_embedded_liberfa():
...
-        erfa_version = erfa.version.erfa_version
-        assert isinstance(erfa_version, str)
```
eerovaher, line 41: If types have to be checked in unit tests then the check should be as strict as possible.

avalentino (approved):

PR #223, opened 2026-04-07
Make `erfa` source directory a constant in `erfa_generator`

eerovaher:
Currently `erfa_generator` accepts the location of the `erfa` source files both as a command line argument and as an argument for its `main()` function, with the default being the correct directory in the `liberfa` submodule of this repository. To me this looks like a leftover from the time before `erfa` was included in `pyerfa` as a submodule and I don't think this functionality is useful. It is much simpler if the location of the `erfa` source files is specified by a module level constant in `erfa_generator`.

This can break backwards compatibility for those running `erfa_generator` directly as a script and for those who import `erfa_generator` and call its `main()` function directly, but I suspect I am the only person who runs the `erfa_generator` script directly and that no-one at all is importing `erfa_generator`.

### DIFF
--- a/erfa_generator.py
+++ b/erfa_generator.py
@@ -15,7 +15,7 @@ import re
 from collections import OrderedDict
 from pathlib import Path
 
-DEFAULT_ERFA_LOC = Path(__file__).with_name("liberfa") / "erfa" / "src"
+ERFA_SRC = Path(__file__).with_name("liberfa") / "erfa" / "src"
 DEFAULT_TEMPLATE_LOC = Path(__file__).with_name("erfa")
 
 NDIMS_REX = re.compile(re.escape("numpy.dtype([('fi0', '.*', <(.*)>)])")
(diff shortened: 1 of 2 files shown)


erfa_generator.py
```diff
@@ -15,7 +15,7 @@
 from collections import OrderedDict
 from pathlib import Path
 
-DEFAULT_ERFA_LOC = Path(__file__).with_name("liberfa") / "erfa" / "src"
+ERFA_SRC = Path(__file__).with_name("liberfa") / "erfa" / "src"
```
eerovaher, line 18: `ERFA_SRC` is the name already used by `setup.py`.

avalentino:
Sorry but I don't see a big advantage in removing the `srcdir` parameter.
I agree that `erfa_fenerator.py` maybe is not used very often but but still IMHO it is a good idea to have its inputs as parameters and configurable configurable via CLI.

Would it be a big issue for you to keep it?

eerovaher:
I don't really like that it is possible to customize the location of the `erfa` source files if `erfa_generator` is invoked directly, but not when `pyerfa` is being installed, and removing functionality from `erfa_generator` is simpler than adding functionality to `setup.py`. But if you really think that this should be configurable from the CLI then close this PR without merging and I will open a new PR that allows configuring the `erfa` directory when installing.

avalentino:
Oh, maybe I made some confusion.
What I meant is that the `erfa_generator.py` should be configurable form CLI so I would not remove the feature from it.
My idea is that someone could need to run it offline to troubleshooting or to develop new functionalities.

Regarding the `setup.py` script , I agree with you that additional configurability is not needed.
We could even invoke the `erfa_generator.py` without parameters from `setup.py`.
The rationale is that the wrapper generation only happens when we generate the packages and it is OK to use the `erfa` C code in the submodule for that.

eerovaher:
Does the location of the templates also have to stay customizable from the CLI?

avalentino:
Yes, to me it makes sense

PR #224, opened 2026-04-08
Remove a Python 3.9 remnant

eerovaher:
I found a remnant of Python 3.9 that wasn't revealed by grepping for neither `3\.9` nor `39`. I've updated the relevant code and also moved it around to make it easier to find when support for Python versions is dropped in the future.


avalentino (approved):
Thanks @eerovaher

PR #227, opened 2026-04-16
Remove a warning about `erfa` not being installed

eerovaher:
The `pyerfa` repository uses flat layout, which means trying to import `erfa` without having installed it is not guaranteed to crash immediately. A warning is meant to inform the user about `erfa` not being installed, but if that failure mode is worth worrying about then it should be addressed properly by using src-layout instead.

For more details see https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/

### DIFF
--- a/erfa/version.py
+++ b/erfa/version.py
@@ -1,15 +1,6 @@
 """Wrapper, ERFA and SOFA version information."""
 
-try:
-    from ._version import version as _version
-except Exception:
-    import warnings
-    warnings.warn(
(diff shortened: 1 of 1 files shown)


avalentino (commented):
This was done to allow people (like me) to build in-place and run unittests from the package root folder.
I find it handy when a package in under heavy development but I agree with you that just raising and exception is safer to ensure that everything is done properly.

PR #233, opened 2026-04-24
Simplify status code templating (and fix two mistakes)

eerovaher:
My goal was to simplify how `erfa_generator` and `core.py.templ` handle status codes, but I also managed to find and fix two mistakes in the process. The difference between the `core.py` generated by the current `main` branch and the `core.py` generated by this PR is
```diff
5365c5365
<     1: 'warning: date outsidethe range 1900-2100 AD',
---
>     1: 'warning: date outside the range 1900-2100 AD',
16298c16298
<     -1: 'illegal case (Note 4)Notes:',
---
>     -1: 'illegal case (Note 4)',
```

The first difference comes from `epv00()`, which has a status code with a two line description. Current `main` merges the lines without any whitespace in between.

The second difference is from `gd2gce()`. In the `erfa` doc comments the section describing status codes is followed by a section called "Note" or "Notes". In the `gd2gce()` doc comment there is no empty line between these sections, so current `main` erroneously merges the title of the next section with the description of the last status code.

### DIFF
--- a/erfa_generator.py
+++ b/erfa_generator.py
@@ -13,6 +13,7 @@ or dtypes for those structs.  They should be added manually in the template file
 import functools
 import re
 from pathlib import Path
+from typing import Final
 
 DEFAULT_ERFA_LOC = Path(__file__).with_name("liberfa") / "erfa" / "src"
 DEFAULT_TEMPLATE_LOC = Path(__file__).with_name("erfa")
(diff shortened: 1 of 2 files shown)


erfa_generator.py
```diff
@@ -263,25 +257,21 @@ def name_for_call(self):
...
+class StatusCodes:
+    def __init__(self, source: str) -> None:
+        self._statuscodes: Final = {
+            "else" if code == "else" else int(code): " ".join(
+                line.strip() for line in description.splitlines()
+            )
+            for code, description in re.findall(
+                r"(-?\w+) = ((?:[^=]+$)+)", source, re.MULTILINE
+            )
+        }
```
eerovaher, line 269: It is worth pointing out that all functions except `epv00()` could be processed with much simpler code:
```suggestion
        self._statuscodes: Final = {
            "else" if code == "else" else int(code): description
            for code, description in re.findall(r"(-?\w+) = (.+)", source)
        }
```

avalentino (approved):

PR #234, opened 2026-04-27
Fix the completely broken `erfa.tpors()` and `erfa.tporv()`

eerovaher:
At least in current `main` and in v2.0.1.6 `erfa_generator` assumes that because `eraTpors()` and `eraTporv()` return an integer then that integer must be a status code, and has therefore generated code in `erfa.tpors()` and `erfa.tporv()` that passes the return value to `check_errwarn()`. But because `erfa_generator` has not been able to find the status code descriptions in the C function doc comments then it has not generated the code for the status code dictionaries `check_errwarn()` requires. Trying to fetch those dictionaries therefore causes an unavoidable `KeyError`, as revealed by the new tests in the first commit. The two functions in question do return an integer, but it is not a status code, so passing it to `check_errwarn()` is the wrong thing to do.

The second commit updates `erfa_generator` so that it no longer assumes that the two C functions return a status code, making the Python wrappers usable. Furthermore, a new check in `erfa_generator` raises an error if it cannot find the status code descriptions in C doc comments it is looking for, which prevents this kind of problem from going unnoticed in the future.

### DIFF
--- a/erfa_generator.py
+++ b/erfa_generator.py
@@ -129,7 +129,7 @@ class ArgumentDoc:
 
 
 class Variable:
-    """Properties shared by Argument and Return."""
+    """Properties shared by Argument, Return and StatusCode."""
     @property
     def npy_type(self):
         """Predefined type used by numpy ufuncs to indicate a given ctype.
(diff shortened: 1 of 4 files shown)


erfa_generator.py
```diff
@@ -257,14 +257,28 @@ def name_for_call(self):
...
+
+        status = re.search(
+            r"Returned \(function value\):\n\s+\w+\s+status.*?:(.+?)\s+Notes?:",
+            doc.doc,
+            re.DOTALL,
+        )
+        if status is None:
+            raise RuntimeError(
+                f"cannot find status code description in {funcname} doc comment"
+            )
```
eerovaher, line 275: I wanted to introduce this check so that if the regex we are currently using does not work for doc comments in a future ERFA release then it would not go unnoticed. I did not expect this error to be raised with two functions in current ERFA. It is clear that checks like this are useful.

avalentino (approved):

avalentino:
Thanks @eerovaher

PR #245, opened 2026-05-08
Remove `Argument.view_dtype` property

eerovaher:
In practice the property has one possible value, so it is simpler to hardcode that value in the `core.py` template.

The full difference between the files generated by `erfa_generator` in current `main` and this PR is
```diff
130d129
< dt_bytes12 = np.dtype('S12')
```

### DIFF
--- a/erfa_generator.py
+++ b/erfa_generator.py
@@ -176,21 +176,6 @@ class Argument(Variable):
         raise ValueError(f"ctype {self.ctype} with shape {self.shape} not recognized.")
 
     @property
-    def view_dtype(self):
-        """Name of dtype corresponding to the ctype for viewing back as array.
-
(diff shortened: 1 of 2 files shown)


erfa_generator.py
```diff
@@ -175,21 +175,6 @@ def dtype(self):
...
-    def view_dtype(self):
-        """Name of dtype corresponding to the ctype for viewing back as array.
-
-        E.g., dt_double for double, dt_double33 for double[3][3].
-
-        The types are defined in core.py, where they are used for view-casts
-        of structured results as regular arrays.
-        """
-        if self.ctype == 'const char':
-            return 'dt_bytes12'
```
eerovaher, line 188: Apparently `self.ctype` is never `"const char"`.

avalentino (approved):

PR #250, opened 2026-05-15
Require `numpy` >= 1.24

eerovaher:
With older `numpy` versions ufuncs without parameters do not work. There are several such functions in `pyerfa`, so those `numpy` versions are not being properly supported anyways. [SPEC 0](https://scientific-python.org/specs/spec-0000/) allows dropping support for `numpy` < 2, so we will still be very generous with our requirements.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,10 +2,12 @@
 ====================
 - Let multi-output functions now return a ``namedtuple`` so that
   the various results can be accessed by attribute. [gh-176]
+- ``pyerfa`` now requires Python 3.10 or later. [gh-185]
 - The ``erfa.tpors()`` and ``erfa.tporv()`` functions no longer raise an
   unavoidable ``KeyError``. [gh-234]
 - ``erfa.cal2jd()`` no longer raises an unexpected ``TypeError`` instead of an
-  ``ErfaError`` or ``ErfaWarning`` if all its inputs are scalars.
+  ``ErfaError`` or ``ErfaWarning`` if all its inputs are scalars. [gh-235]
+- ``pyerfa`` now requires ``numpy`` 1.24 or later. [gh-250]
 
 2.0.1.6 (2025-01-27)
 ====================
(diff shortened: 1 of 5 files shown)


avalentino (approved):

mhvk (commented):
Thanks, this is indeed overdue. Only one request: could you add an entry in `CHANGES.rst`? Those entries will be put in the release notes (and really we should have a release soon, so this is another good reason to do so).

CHANGES.rst
```diff
@@ -2,10 +2,12 @@
 ====================
 - Let multi-output functions now return a ``namedtuple`` so that
   the various results can be accessed by attribute. [gh-176]
+- ``pyerfa`` now requires Python 3.10 or later. [gh-185]
```
eerovaher, line 5: In the current change log `pyerfa` refers to itself in at least three different ways:
https://github.com/liberfa/pyerfa/blob/aa06b0635a822c4dd903efe5c4e96119b039826b/CHANGES.rst?plain=1#L48
https://github.com/liberfa/pyerfa/blob/aa06b0635a822c4dd903efe5c4e96119b039826b/CHANGES.rst?plain=1#L37
https://github.com/liberfa/pyerfa/blob/aa06b0635a822c4dd903efe5c4e96119b039826b/CHANGES.rst?plain=1#L13

I chose to use ``pyerfa`` because that is what has been used most recently.

mhvk: Wasn't this PR about numpy? In itself fine to mention python too!

eerovaher: This PR only updates the `numpy` requirement, but I noticed that I had forgotten to add the PR number to the ``erfa.cal2jd()`` bugfix entry, and then I thought if I am updating the change log I might as well add the entry for the Python version requirement too.

mhvk: Ah, sorry, I had missed the numpy addition further down, so all is OK.

mhvk (approved):

PR #251, opened 2026-05-18
Start enforcing the Ruff RUF rules

eerovaher:
I am configuring Ruff to enforce its [RUF rules](https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf). There were two kinds of RUF violations that had to be fixed.

`docs/conf.py` contained non-ASCII characters that are visually indistinguishable or almost indistinguishable from ASCII characters. Personally I am not opposed to using non-ASCII characters when they improve readability (e.g. writing `α` instead of `alpha`), but visually indistinguishable characters cannot possibly do that.

In `erfa/tests/test_erfa.py` Ruff pointed out several unused unpacked variables. If a `pyerfa` Python wrapper function returns a `tuple` it will have named attributes, so if only one or two values are needed then unpacking the tuple does not improve readability. I've rewritten the test in question as two separate parametrized tests.

### DIFF
--- a/erfa/tests/test_erfa.py
+++ b/erfa/tests/test_erfa.py
@@ -60,40 +60,43 @@ def test_version_with_embedded_liberfa():
     assert erfa.__version__.startswith(erfa.version.erfa_version)
 
 
-def test_erfa_wrapper():
-    """
-    Runs a set of tests that mostly make sure vectorization is
-    working as expected
-    """
+@pytest.mark.parametrize(
+    ("ra", "dec", "jd", "shape"),
+    [
+        pytest.param(0.0, 0.0, 2456855.5, (), id="scalar"),
+        pytest.param(
+            0.0, 0.0, np.linspace(2456855.5, 2456856.5, 5), (5,), id="simple array"
+        ),
+        pytest.param(
+            np.linspace(0.0, np.pi * 2.0, 5)[:, None, None],
+            np.linspace(-np.pi / 2.0, np.pi / 2.0, 4)[:, None],
+            np.linspace(2456855.5, 2456856.5, 3),
+            (5, 4, 3),
+            id="broadcasting",
+        ),
+    ],
+)
+def test_atco13_shape(ra, dec, jd, shape):
+    assert (
+        erfa.atco13(
+            ra, dec, 0, 0, 0, 0, jd, 0, 0, 0, np.pi / 4, 0, 0, 0, 1014, 0, 0, 0.5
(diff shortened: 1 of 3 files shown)


erfa/tests/test_erfa.py
```diff
@@ -60,40 +60,43 @@ def test_version_with_embedded_liberfa():
...
-
-    aob, zob, hob, dob, rob, eo = erfa.atco13(
-        0.0, 0.0, 0.0, 0.0, 0.0, 0.0, jd[0],
-        0.0, 0.0, 0.0, np.pi/4.0, 0.0, 0.0, 0.0, 1014.0, 0.0, 0.0, 0.5)
-    assert aob.shape == ()
-
-    aob, zob, hob, dob, rob, eo = erfa.atco13(
-        ra[:, None, None], dec[None, :, None], 0.0, 0.0, 0.0, 0.0, jd[None, None, :],
-        0.0, 0.0, 0.0, np.pi/4.0, 0.0, 0.0, 0.0, 1014.0, 0.0, 0.0, 0.5)
-    (aob.shape) == (5, 4, 121)
```
eerovaher, line 86: There is no `assert` here, so this was a useless comparison. The Ruff rule [B015 (useless-comparison)](https://docs.astral.sh/ruff/rules/useless-comparison/) would guard against mistakes like that, but we haven't turned on the B rules yet.

erfa/tests/test_erfa.py
```diff
@@ -60,40 +60,43 @@ def test_version_with_embedded_liberfa():
...
-    """
-    Runs a set of tests that mostly make sure vectorization is
-    working as expected
-    """
+@pytest.mark.parametrize(
+    ("ra", "dec", "jd", "shape"),
+    [
+        pytest.param(0.0, 0.0, 2456855.5, (), id="scalar"),
+        pytest.param(
+            0.0, 0.0, np.linspace(2456855.5, 2456856.5, 5), (5,), id="simple array"
```
avalentino, line 68: This does not look like the original one
Do I miss something?

eerovaher: I made the arrays smaller because I didn't see the point in forcing the tests to run with larger arrays, but the tests are still checking that the output shape is correct. I did also change the range of `jd` values to shorten setup, but that shouldn't matter because with or without my patch the asserts here only care about the shape of the outputs, not about the values within.

avalentino (approved):

PR #260, opened 2026-06-01
Remove the `Function.python_call` property

eerovaher:
`Function.python_call` generates ufunc calls for the Python wrappers and splits the generated code over multiple lines if it is too long for a single line. However, the automatically generated `erfa/core.py` file is not meant to be edited by humans and reading it is not informative because the functions there are just thin wrappers around the corresponding ufuncs. Therefore, having code in `erfa_generator` to format the generated code (in practice 8 functions) is not worth the effort. Even if formatting the generated code is considered important then it would be better to use a dedicated formatter such as [Black](https://black.readthedocs.io/en/stable/) or the [Ruff formatter](https://docs.astral.sh/ruff/formatter/) when a release is being made.

### DIFF
--- a/erfa_generator.py
+++ b/erfa_generator.py
@@ -333,25 +333,16 @@ class Function:
              for args in (self.args_by_inout('in|inout'),
                           self.args_by_inout('inout|out|ret|stat'))])
 
-    @property
-    def python_call(self):
-        result = _assemble_py_func_call(
-            "ufunc." + self.pyname,
-            in_args=[arg.name for arg in self.args_by_inout("in|inout")],
-            out_args=[arg.name for arg in self.args_by_inout("inout|out|stat|ret")],
-        )
-        if len(result) < 75:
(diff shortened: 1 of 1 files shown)


avalentino:
Hi @eerovaher, do you think that it would make sense to add a call to `black` in this PR?
It could be in `erfa_geenrator.py` itself or in `setup.py`.

eerovaher:
Formatting the generated files right now would change them quite a lot. For example, the generated `erfa/core.py` uses both `'` and `"` in string literals, but formatters prefer the latter. My preference would be not to automatically format any generated files until the differences between unformatted and formatted files are as small as possible without needing any formatting-specific conditional branching in `erfa_generator`. I can work on that tomorrow. If this PR is still open when I'm done then I can expand its scope, or I can open a separate PR if this one gets merged before that.

avalentino (approved):

PR #263, opened 2026-06-03
Avoid initializing output arguments in translated tests

eerovaher:
Among other things, `erfa_generator` translates the ERFA tests in `t_erfa_c.c` into a `pytest` test suite for `pyerfa` ufuncs. The C tests have to initialize output arguments (or at least allocate them), but in the Python tests that step can be skipped without losing any test coverage. This PR shortens the generated `erfa/tests/test_ufunc.py` from 3764 lines to 3617.

### DIFF
--- a/erfa_generator.py
+++ b/erfa_generator.py
@@ -413,8 +413,11 @@ class TestFunction:
                     defines.append(var)
                 # Is variable an array?
                 name, _, rest = var.partition("[")
-                # If not, or one of iymdf or ihmsf, ignore (latter are outputs only).
-                if not rest or rest[:2] == "4]":
+                if (
+                    not rest
+                    or name in self.func.doc.output  # no need to initialize outputs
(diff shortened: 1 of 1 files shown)


erfa_generator.py
```diff
@@ -413,8 +413,11 @@ def process_definitions(self) -> list[str]:
                     defines.append(var)
                 # Is variable an array?
                 name, _, rest = var.partition("[")
-                # If not, or one of iymdf or ihmsf, ignore (latter are outputs only).
-                if not rest or rest[:2] == "4]":
+                if (
+                    not rest
+                    or name in self.func.doc.output  # no need to initialize outputs
+                    or name == "iydmf"  # eraJdcalf test has a typo
```
eerovaher, line 419: The typo in the `eraJdcalf` test also means that `iymdf` the old comment refers to cannot be found by grepping.

avalentino (approved):

PR #265, opened 2026-06-04
Implement the `Function.c_retval` attribute

eerovaher:
Currently the code in `erfa_generator` and the templates that looks at the return value of ERFA C functions calls `Function.args_by_inout()`, but that always returns a `list`, which does not help convey that there can only be at most one status code or return value. The new `Function.c_retval` attribute makes that fact explicit and allows replacing several for-loops in the templates with more natural if-statements.

### DIFF
--- a/erfa/ufunc.c.templ
+++ b/erfa/ufunc.c.templ
@@ -200,9 +200,9 @@ static void ufunc_loop_{{ func.pyname }}(
    {%- endif %}
   {%- endfor %}
   {#- /* variables to hold status and return values */ #}
-  {%- for arg in func.args_by_inout('stat|ret') %}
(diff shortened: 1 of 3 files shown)


erfa/ufunc.c.templ
```diff
@@ -321,18 +321,14 @@ static void ufunc_loop_{{ func.pyname }}(
       {#- /*
            * call the actual erfa function
            */ #}
-        {{ func.args_by_inout('ret|stat') |
-                    map(attribute='name') |
-                    surround('_', ' = ') |
```
eerovaher, line 326: This was the last place that used the `surround` filter defined in `erfa_generator.main()`.

avalentino (approved):

PR #267, opened 2026-06-08
Replace the only Jinja macro with Python code

eerovaher:
I am not aware of any good reasons why code should be written as Jinja macros rather than Python functions. Luckily it is not very complicated to get rid of the macros because there is only one.

### DIFF
--- a/erfa_generator.py
+++ b/erfa_generator.py
@@ -196,6 +196,23 @@ class Argument(Variable):
                 return "(3, 3)"
         return super().signature_shape
 
+    def inner_loop_steps_and_copy(self, name_suffix: str = "") -> str | None:
+        if self.signature_shape == "()":
+            return None
+        name = self.name + name_suffix
+        lines = [f"npy_intp is_{name}{i} = *steps++;" for i in range(self.ndim or 1)]
+        # copy should be made if buffer not contiguous;
+        # note: one can only have 1 or 2 dimensions
+        lines.append(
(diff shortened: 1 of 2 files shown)


erfa_generator.py
```diff
@@ -196,6 +196,23 @@ def signature_shape(self):
...
+    def inner_loop_steps_and_copy(self, name_suffix: str = "") -> str | None:
+        if self.signature_shape == "()":
+            return None
+        name = self.name + name_suffix
+        lines = [f"npy_intp is_{name}{i} = *steps++;" for i in range(self.ndim or 1)]
+        # copy should be made if buffer not contiguous;
+        # note: one can only have 1 or 2 dimensions
+        lines.append(
+            f"int copy_{name} = (is_{name}0 != sizeof({self.ctype}));"
+            if self.ndim <= 1
```
eerovaher, line 208: Based on the comment I would have expected that we could check `if self.ndim == 1`, but turns out that it is necessary to use `<=` instead of `==` because in a few cases `self.ndim` is `0`. Furthermore, in the templates the `Argument.ndim` property is always used as `arg.ndim or 1`, and this function also uses the analogous `self.ndim or 1`. I am getting the impression that something is being set up wrong, and instead of fixing the problem where it occurs the current code contains workarounds wherever the problem is encountered. I'll have to investigate what the real problem is and how to solve it.

avalentino (approved):

PR #268, opened 2026-06-09
Implement the `Argument.cast_pointer` property

eerovaher:
The new property helps avoid code repetition in `erfa/ufunc.c.templ`.

I don't know C, so I'm not really sure what the generated C code is supposed to be doing and therefore it is possible the name I gave to the property is not good. If there is a better name then do let me know.


avalentino (approved):
I'm not too convinced by the name `cast_pointer` but also I do not have anything meaningful to suggest.
I will just merge

PR #269, opened 2026-06-11
Improve generating code for arrays with shapes not known at compile time

eerovaher:
Whenever the `Argument.ndim` property is accessed in `erfa_generator` or in the ufuncs template it is used as `ndim or 1` (or something equivalent), meaning that `Argument.ndim` is reporting some one-dimensional arrays to be scalars instead. The real problem is that the `shape` attribute of an array with size not known at compile-time is set to an empty tuple, even if the number of dimensions is known and non-zero.

The first commit here implements the `Argument.copy_elements()` method, which reduces code duplication in the ufuncs template. In particular, it de-duplicates the code that would have to be edited anyways.

The second commit changes the type of the `Argument.shape` attribute from `tuple[int, ...]` to `tuple[int | None, ...]`, where `None` represents a value not known at compile-time. This fixes the `Argument.ndim` property without having to change the property's implementation.

### DIFF
--- a/erfa_generator.py
+++ b/erfa_generator.py
@@ -100,14 +100,11 @@ class Argument(Variable):
     def __init__(self, definition: str, doc: FunctionDoc) -> None:
         self.doc = doc
         ctype, ptr_name_arr = definition.strip().rsplit(" ", 1)
-        name_arr = ptr_name_arr.removeprefix("*").removesuffix("[]")
-        self.is_ptr = name_arr != ptr_name_arr
-        if "[" in name_arr:
-            name_arr, arr = name_arr.split("[", 1)
-            self.shape = tuple([int(size) for size in arr[:-1].split("][")])
-        else:
-            self.shape = ()
-        super().__init__(ctype, name_arr)
+        self.is_ptr: Final = ptr_name_arr.startswith("*")
+        self.shape: Final = tuple(
+            int(s) if s else None for s in re.findall(r"\[(\d*)\]", ptr_name_arr)
+        )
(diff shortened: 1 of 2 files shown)


erfa_generator.py
```diff
@@ -171,19 +168,26 @@ def dtype(self):
...
     @property
-    def cshape(self):
-        return ''.join([f'[{s}]' for s in self.shape])
+    def cshape(self) -> str:
+        elems = []
+        for s in self.shape:
+            if s is None:
+                return ""
+            elems.append(f"[{s}]")
+        return "".join(elems)
```
eerovaher, line 190: I'm not convinced the new implementation of `Argument.cshape` is the best possible, but it is good enough to avoid any changes in the generated `erfa/ufuncs.c` file.

erfa_generator.py
```diff
@@ -200,12 +204,12 @@ def inner_loop_steps_and_copy(self, name_suffix: str = "") -> str | None:
         if self.signature_shape == "()":
             return None
         name = self.name + name_suffix
-        lines = [f"npy_intp is_{name}{i} = *steps++;" for i in range(self.ndim or 1)]
+        lines = [f"npy_intp is_{name}{i} = *steps++;" for i in range(self.ndim)]
         # copy should be made if buffer not contiguous;
         # note: one can only have 1 or 2 dimensions
         lines.append(
             f"int copy_{name} = (is_{name}0 != sizeof({self.ctype}));"
-            if self.ndim <= 1
```
eerovaher, line 208: The old code could have been written as `if (self.ndim or 1) == 1`.

avalentino (approved):
Thanks @eerovaher

PR #271, opened 2026-06-15
Rework argument role tracking in code generation

eerovaher:
Some ERFA function arguments are inputs to the function, others are pointers to where the output is supposed to be written, and a few are (pointers to) structs with some fields being function inputs and others locations for the output. The code generation in `erfa_generator` and the templates has to keep track of the role of each argument. If the ERFA function returns a status code or a result value then that has to be tracked too. Currently this is done by the `Function.args_by_inout()` method, which checks the role of each argument every time it is called, but that has several shortcomings. For example, both `Return` and `StatusCode` have an `inout_state` attribute which is redundant because it is fully determined by the class, but it is nonetheless needed to simplify `Function.args_by_inout()` implementation.

The new approach is to replace the single `Function.args_by_inout()` method that can produce different lists based on its argument with a few different `Function` attributes and properties. In practice the new approach is simpler to work with. For example, IDEs can much more easily suggest completions for attributes and properties than for the `Function.args_by_inout()` argument.

There are no changes to any of the files `erfa_generator` creates.

### DIFF
--- a/erfa_generator.py
+++ b/erfa_generator.py
@@ -106,16 +106,6 @@ class Argument(Variable):
         )
         super().__init__(ctype, ptr_name_arr.removeprefix("*").split("[", 1)[0])
 
-    @functools.cached_property
-    def inout_state(self) -> str:
-        if self.name in self.doc.input:
-            return "in"
-        if self.name in self.doc.inout:
-            return "inout"
-        if self.name in self.doc.output:
-            return "out"
-        return ""
-
     @property
     def name_for_call(self) -> str:
         """How the argument should be used in the call to the ERFA function.
@@ -231,7 +221,6 @@ class Argument(Variable):
 
 class StatusCode(Variable):
     def __init__(self, ctype: str, doc: FunctionDoc, funcname: str) -> None:
-        self.inout_state = "stat"
         super().__init__(ctype)
 
         status = re.search(
@@ -260,10 +249,7 @@ class StatusCode(Variable):
 
 
 class Return(Variable):
-
-    def __init__(self, ctype: str) -> None:
-        self.inout_state = "ret"
-        super().__init__(ctype)
+    pass
 
 
 class ResultTuple:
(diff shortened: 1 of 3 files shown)


erfa_generator.py
```diff
@@ -304,9 +290,16 @@ def __init__(self, name: str, source_path: Path) -> None:
...
-            Argument(arg, self.doc) for arg in re.findall("[^,]+", search.group(2))
-        ]
+        args = [Argument(arg, self.doc) for arg in re.findall("[^,]+", search.group(2))]
+        self.in_args: Final = tuple(a for a in args if a.name in self.doc.input)
+        self.inout_args: Final = tuple(a for a in args if a.name in self.doc.inout)
+        self.out_args: Final = tuple(a for a in args if a.name in self.doc.output)
+
+        self.py_args: Final = (*self.in_args, *self.inout_args)
+        self.c_args: Final = (*self.py_args, *self.out_args)
+        self.inout_or_out_args: Final = (*self.inout_args, *self.out_args)
```
eerovaher, line 300: I chose to implement an attribute if I could fit it in a single line and a property if it would have taken several lines. This keeps the `Function.__init__()` method reasonably short while avoiding having to implement several tiny methods in `Function`.

erfa_generator.py
```diff
@@ -315,30 +308,29 @@ def __init__(self, name: str, source_path: Path) -> None:
             )
             self.args.append(self.c_retval)
 
-        self.result_tuple: Final = (
-            ResultTuple(self.pyname, py_return)
-            if len(py_return := self.args_by_inout("inout|out|ret")) > 1
+    @functools.cached_property
```
eerovaher, line 311: The main reason for using `@functools.cached_property` instead of `@property` is to help convey that the implementation of the property is a [pure function](https://en.wikipedia.org/wiki/Pure_function).

avalentino (approved):

erfa_generator.py
```diff
@@ -304,9 +290,16 @@ def __init__(self, name: str, source_path: Path) -> None:
...
         self.doc: Final = FunctionDoc(search.group(3), self.pyname)
-        self.args: Final[list[Variable]] = [
-            Argument(arg, self.doc) for arg in re.findall("[^,]+", search.group(2))
-        ]
+        args = [Argument(arg, self.doc) for arg in re.findall("[^,]+", search.group(2))]
+        self.in_args: Final = tuple(a for a in args if a.name in self.doc.input)
+        self.inout_args: Final = tuple(a for a in args if a.name in self.doc.inout)
+        self.out_args: Final = tuple(a for a in args if a.name in self.doc.output)
+
+        self.py_args: Final = (*self.in_args, *self.inout_args)
```
eerovaher, line 298: I wasn't sure if this attribute should be called `py_args` or `ufunc_args`. It might even be a good idea to define `ufunc_args` as an alias. The main advantage of `py_args` is that it is shorter.

PR #273, opened 2026-06-17
WHL: drop win32 support (stop distributing wheels)

neutrinoceros:
Following numpy
https://mail.python.org/archives/list/numpy-discussion@python.org/thread/FGVCDS5JICVIUVBISGWZNQ46QZ3XX5PB/#TTSOQQ3XLXXK52CZYMN7MALJQEI65UDN

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -47,7 +47,6 @@ jobs:
         - pp3*-macosx_x86_64
         - pp3*-macosx_arm64
         # Windows wheels
-        - cp3*-win32
         - cp3*-win_amd64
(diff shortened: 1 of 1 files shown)


avalentino:
Dear @neutrinoceros 
It would be fine for me.
Just I don't know if this can have an impact on astropy.
I guess that we should drop the architecture in a coordinated way (although there is always the possibility to use win-32 wheels that are currently published for the stable pyerfa).

neutrinoceros:
I opened the same patch at https://github.com/astropy/astropy/pull/19931

avalentino (approved):

PR #279, opened 2026-06-25
Move code from `ufunc.c.templ` to `erfa_generator.py`

eerovaher:
The new Python code helps avoid code repetition in the `erfa/ufunc.c` template. Furthermore, Python code can be checked with tools such as Ruff or Mypy and there is better IDE support. If we keep moving code from the templates to `erfa_generator.py` then it might eventually become possible to drop `jinja2` as a build requirement.

### DIFF
--- a/erfa/ufunc.c.templ
+++ b/erfa/ufunc.c.templ
@@ -220,12 +220,7 @@ static void ufunc_loop_{{ func.pyname }}(
        * Actual inner loop, increasing all pointers by their steps
        */ #}
     for (i_o = 0; i_o < n_o;
-         i_o++ {%- for arg in func.args -%}
-             , {{ arg.name }} += s_{{ arg.name }}
-             {%- endfor -%}
-             {%- for arg in func.inout_args -%}
-             , {{ arg.name }}_in += s_{{ arg.name }}_in
-             {%- endfor -%}) {
+         i_o++, {{ func.increment_arg_pointers }}) {
       {%- if func.signature %}
        {#- /*
             * GENERALIZED UFUNC, prepare for call.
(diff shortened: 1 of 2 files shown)


erfa/ufunc.c.templ
```diff
@@ -220,24 +220,17 @@ static void ufunc_loop_{{ func.pyname }}(
...
        {#- /* copy input arguments to buffer if needed */ #}
        {%- for arg in func.in_args %}
         {%- if arg.signature_shape != '()' %}
-        if (copy_{{ arg.name }}) {
-            {{ arg.copy_elements("to") }}
-        }
+        {{ arg.cast_pointer_if_needed | indent(8) }}
         else {
-            {{ arg.cast_pointer }}
+            {{ arg.copy_elements("to") }}
```
eerovaher, line 233: I've changed the order of the branches here, which does introduce differences in the generated C code, but there should be no changes in what the code does.

erfa_generator.py
```diff
@@ -202,6 +202,16 @@ def inner_loop_steps_and_copy(self, name_suffix: str = "") -> str | None:
...
 
+    @functools.cached_property
+    def cast_pointer_if_needed(self) -> str:
+        return "\n".join(
+            [
+                f"if (!copy_{self.name}) {{",
+                f"    {self.cast_pointer}",
+                "}",
+            ]
+        )
```
eerovaher, line 213: I've been formatting the code I touch (including this function) using the Ruff formatter with its default style. With the [preview style](https://docs.astral.sh/ruff/formatter/#preview-style) this function would look better:
```python
     @functools.cached_property
     def cast_pointer_if_needed(self) -> str:
         return "\n".join([
             f"if (!copy_{self.name}) {{",
             f"    {self.cast_pointer}",
             "}",
         ])
```
If we can agree that we like the preview style more then I could turn that on.

avalentino: OK for me

avalentino (approved):

PR #285, opened 2026-07-02
Move some gufunc-specific code from `ufunc.c` template to `GUFunc` class

eerovaher:
When I implemented the `GUFunc` class in 489b9a87bccfef8e22a7a1040c9a06152f3e6ce6 I made a simplifying assumption that code which is specific to gufuncs could be found by grepping for `func.signature` in the `ufunc.c` template, but now I've found some gufunc code that did not mention `func.signature` at all. Such code should be implemented in the `GUFunc` class, not in any template.

### DIFF
--- a/erfa/ufunc.c.templ
+++ b/erfa/ufunc.c.templ
@@ -152,19 +152,7 @@ static inline void copy_to_eraLDBODY(char *ptr, npy_intp s, npy_intp n,
 static void ufunc_loop_{{ func.pyname }}(
     char **args, npy_intp const *dimensions, npy_intp const* steps, void* data)
 {
-    {{ func.init_ufunc_loop_local_vars | indent(4) }}
-  {#- /*
-       * Actual inner loop, increasing all pointers by their steps
-       */ #}
-    for (i_o = 0; i_o < n_o;
-         i_o++, {{ func.increment_arg_pointers }}) {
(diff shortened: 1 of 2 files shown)


erfa/ufunc.c.templ
```diff
@@ -152,19 +152,7 @@ static inline void copy_to_eraLDBODY(char *ptr, npy_intp s, npy_intp n,
...
 {
-    {{ func.init_ufunc_loop_local_vars | indent(4) }}
-  {#- /*
-       * Actual inner loop, increasing all pointers by their steps
-       */ #}
-    for (i_o = 0; i_o < n_o;
-         i_o++, {{ func.increment_arg_pointers }}) {
-        {{ func.ufunc_loop_inner_loop_body | indent(8) }}
-    }
-    {%- if func.user_dtype == 'dt_eraLDBODY' %}
```
eerovaher, line 163: If `func.user_dtype == 'dt_eraLDBODY'` then that means there must be an `Argument` with `ctype` value `eraLDBODY`, but that in turn means the `signature_shape` of that `Argument` is `"(n)"`, not `"()"`, so `Function.from_c_code()` indeed decides to create a `GUFunc`, not a `UFunc`.

avalentino (approved):

PR #292, opened 2026-07-09
Add separate templates for `ufunc_loop_*`

eerovaher:
There are now separate `ufunc_loop_*` templates for functions with `user_dtype=="dt_eraLDBODY"` and functions with any other `user_dtype`, which simplifies the implementation of `erfa_generator`. The big difference between the new templates and the templates used with `jinja` is that the new templates do not contain any conditional branching nor any loops, they only contain placeholders that `erfa_generator` replaces.

### DIFF
--- a/erfa_generator.py
+++ b/erfa_generator.py
@@ -316,11 +316,13 @@ class Function(ABC):
         doc: FunctionDoc,
         args: Sequence[Argument],
         c_retval: Return | StatusCode | None,
+        templateloc: Path,
     ) -> None:
         self.name: Final = name
         self.pyname: Final = name.removeprefix("era").lower()
         self.doc: Final = doc
         self.c_retval: Final = c_retval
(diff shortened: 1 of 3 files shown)


erfa_generator.py
```diff
@@ -448,18 +448,21 @@ def init_ufunc_loop_local_vars(self) -> str:
         return "\n".join(lines)
 
     @functools.cached_property
-    def ufunc_loop_body(self) -> str:
+    def ufunc_loop_template(self) -> Template:
+        return Template((self.templateloc / "ufunc_loop.templ").read_text().strip())
```
eerovaher, line 452: This property means a template file is read in for every ERFA function that is rendered. My preference would be to store the contents of the template files in a global constant so that every template file is read in just once, but it is not possible to use global constants for that as long as `main()` has the `templateloc` parameter.

avalentino (approved):

PR #297, opened 2026-07-16
Drop `jinja2` as a build dependency

eerovaher:
In this PR the templates that were rendered with `jinja2` are now rendered using only the Python standard library and `erfa_generator.py` (`erfa/tests/test_ufunc.py.templ` is updated in this PR, others have already been updated earlier). This has required moving all conditional branching and loops from the templates to `erfa_generator.py`, but that was a good tradeoff because Python has better IDE support and the code in `erfa_generator.py` can be checked with tools such as mypy.

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -141,7 +141,7 @@ jobs:
         python-version: '3.10'
     - name: Install build-time dependencies
       run: |
-        python -m pip install --upgrade wheel setuptools setuptools_scm jinja2 'numpy<2.0'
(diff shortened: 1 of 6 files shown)


avalentino (approved):

mhvk (approved):
I had not seen the previous PR, but thanks very much for doing this -- it has been on my internal wish list for a while to get rid of jinja2 since I don't like having another "language" to deal with. 

Possibly we can even get rid of the templates themselves, but since that would really just entail having a large string with their contents in the python files, there probably is no need.

Again, really nice!

PR #298, opened 2026-07-17
Simplify rendering the templates

eerovaher:
This PR simplifies the `erfa_generator` `main()` function by moving repetitive code to a new `_render_template()` function. This does add a newline `\n` to the end of the rendered `erfa/ufunc.c` file, but there's no good reason why `erfa/ufunc.c` shouldn't end with a `\n`, so it is a good change.

### DIFF
--- a/erfa_generator.py
+++ b/erfa_generator.py
@@ -859,6 +859,12 @@ def _docstring_section_title(title: str) -> tuple[str, str, str]:
     return ("", title, len(title) * "-")
 
 
+def _render_template(template: Path, /, **kwargs: str) -> None:
+    template.with_suffix("").write_text(
+        Template(template.read_text()).substitute(**kwargs)
+    )
+
+
 def main(srcdir: Path, templateloc: Path) -> None:
     funcs = [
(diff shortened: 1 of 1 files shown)


erfa_generator.py
```diff
@@ -879,54 +885,40 @@ def main(srcdir: Path, templateloc: Path) -> None:
...
+                *[f'"{func.pyname}",' for func in funcs],
+            ])
+        ),
+        constants="\n".join(constant.define for constant in constants),
+        funcs="\n\n\n".join([func.to_python for func in funcs]),
     )
 
-    ufuncfn = "ufunc.c"
-    (templateloc / ufuncfn).write_text(
-        Template((templateloc / f"{ufuncfn}.templ").read_text().strip()).substitute(
```
eerovaher, line 900: I wanted to avoid any changes to the generated files in the PRs that switched rendering from `jinja` to Python standard library so that it would be simple to check if I had made any mistakes in the process. For `erfa/ufunc.c` that meant I had to explicitly remove the trailing `\n` with the `.strip()` call.

avalentino (approved):

PR #301, opened 2026-07-22
Simplify `setup.py`

eerovaher:
There are a handful of simplifications that can be applied to `setup.py` without changing what it does.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -102,7 +102,7 @@ elif not gen_files_exist:
 
 if not gen_files_exist or any(fn.stat().st_mtime < src_mtime for fn in GEN_FILES):
     print('Run "erfa_generator.py"')
-    subprocess.run([sys.executable, "erfa_generator.py", ERFA_SRC], check=True)
+    subprocess.run([sys.executable, "erfa_generator.py"], check=True)
 
 sources = [Path("erfa", "ufunc.c")]
 include_dirs = [np.get_include()]
@@ -116,7 +116,6 @@ else:
         for file in ERFA_SRC.iterdir()
         if file.suffix == ".c" and not file.name.startswith("t_")
     )
-    include_dirs.append(str(ERFA_SRC))
     # liberfa configuration
     config_h = LIBERFADIR / "config.h"
     if not config_h.exists():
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -137,16 +136,11 @@ def guess_next_dev(version: ScmVersion) -> str:
                 config_h.write_text(version_definitions)
             else:
                 warn("unable to get liberfa version")
-    if config_h.exists():
```
eerovaher, line 140: As far as I can tell the `if`-statement should have been removed when 0148ea07deef97aacf343db0d41b166d367c48e4 removed the accompanying `elif`-block.

avalentino: Looking at 0148ea0, the problem is that without a `config.h` it is not possible to generate a valid `sdist`.
I think (not sure by the way) that it is still possible to build `liberfa` without `config.h` but in that case `HAVE_CONFIG_H` should not be set.

IMHO the old behavior (pre 0148ea0) was correct.

If you still want to simplify then at least the warning above should become a `raise SomeException(...)`, maybe.

setup.py
```diff
@@ -137,16 +136,11 @@ def guess_next_dev(version: ScmVersion) -> str:
...
-        include_dirs.append(str(LIBERFADIR))
-        define_macros.append(("HAVE_CONFIG_H", "1"))
-
-use_scm_version = {
-    'version_scheme': guess_next_dev,
-}
+    include_dirs.extend(map(str, [ERFA_SRC, LIBERFADIR]))
+    define_macros.append(("HAVE_CONFIG_H", "1"))
```
avalentino, line 141: This is not OK if the user wants to use the system version of liberfa (`PYERFA_USE_SYSTEM_LIBERFA` set in the environment).
The last two instructions should be part of the `else` clause.

avalentino (changes requested):

avalentino:
Apparently these is some issue with the CI.
It seems unrelated so I will merge anyways
@eerovaher do you have any idea about what could be the issue?

eerovaher:
From what I can tell the CI failure is not caused by us. We have to wait for OpenAstronomy to fix the problem.

PR #309, opened 2026-08-03
Manual monthly `pre-commit` update

eerovaher:
The automatic monthly `pre-commit` update (#308) failed because [Ruff 0.16 changed which lint rules are enabled by default](https://astral.sh/blog/ruff-v0.16.0#better-default-rule-set). I added a few (temporary) ignores to the Ruff configuration so that `pre-commit` checks could succeed.

Closes #308

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -18,7 +18,7 @@ import time
 from datetime import datetime
 
 try:
-    from sphinx_astropy.conf.v1 import *  # noqa: F403
+    from sphinx_astropy.conf.v1 import *
     from sphinx_astropy.conf.v1 import exclude_patterns
 except ImportError:
     print('ERROR: the documentation requires the sphinx-astropy package to be installed')
--- a/setup.py
+++ b/setup.py
@@ -101,7 +101,7 @@ if not BUILDING_FROM_SDIST:
 sources = [Path("erfa", "ufunc.c")]
 include_dirs = [np.get_include()]
 libraries = []
-if int(os.environ.get("PYERFA_USE_SYSTEM_LIBERFA", 0)):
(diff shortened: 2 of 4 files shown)


docs/conf.py
```diff
@@ -18,7 +18,7 @@
 from datetime import datetime
 
 try:
-    from sphinx_astropy.conf.v1 import *  # noqa: F403
+    from sphinx_astropy.conf.v1 import *
```
eerovaher, line 21: F403 is no longer enforced by default.

setup.py
```diff
@@ -101,7 +101,7 @@ def guess_next_dev(version: ScmVersion) -> str:
 sources = [Path("erfa", "ufunc.c")]
 include_dirs = [np.get_include()]
 libraries = []
-if int(os.environ.get("PYERFA_USE_SYSTEM_LIBERFA", 0)):
+if int(os.environ.get("PYERFA_USE_SYSTEM_LIBERFA", "0")):
```
eerovaher, line 104: [PLW1508 (invalid-envvar-default)](https://docs.astral.sh/ruff/rules/invalid-envvar-default/) is now turned on by default, but it was simpler to enforce it than to configure Ruff to ignore it.

avalentino (approved):

avalentino:
Thanks a lot @eerovaher 
In the last period you are the most active contributor.
If you are interested, an everybody in @liberfa/pyerfa-maintainers agrees, I would be happy to give you commit access to the repository.

eerovaher:
It is not clear to me what I would benefit from having write access to this repository because someone else would still have to review my PRs. However, triage permissions would be useful because then I could add labels to my PRs myself.

avalentino:
done

PR #315, opened 2026-08-11
Make `erfa_generator._indent()` more convenient

eerovaher:
On current `main` `erfa_generator._indent()` works on multiline strings just like the [`jinja` `indent()` filter](https://jinja.palletsprojects.com/en/stable/templates/#jinja-filters.indent) the function was meant to replace, but it is more convenient if it works on a `list` of strings, where each element represents a single line, because that avoids having to call `"\n".join()` over and over again.

There are no changes to the files `erfa_generator` creates.

### DIFF
--- a/erfa_generator.py
+++ b/erfa_generator.py
@@ -12,6 +12,7 @@ import re
 import textwrap
 from abc import ABC, abstractproperty
 from collections.abc import Iterable, Sequence
+from itertools import chain
 from pathlib import Path
 from string import Template
 from typing import Final, final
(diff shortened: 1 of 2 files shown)


erfa_generator.py
```diff
@@ -833,9 +834,11 @@ def main(srcdir: Path, templateloc: Path) -> None:
         templateloc / "ufunc.c.templ",
         ufunc_loops="\n\n".join(func.ufunc_loop for func in funcs),
         type_and_func_definitions=_indent(
-            "\n".join(filter(None, [func.define_types_and_functions for func in funcs]))
+            list(chain(*[func.define_types_and_functions for func in funcs]))
+        ),
+        ufunc_definitions=_indent(
+            list(chain(*[func.define_ufunc.splitlines() for func in funcs]))
         ),
```
eerovaher, line 841: When `pyerfa` drops support for Python 3.14 then [PEP 798 – Unpacking in Comprehensions](https://peps.python.org/pep-0798/) will allow this to be rewritten as
```suggestion
            [*func.define_types_and_functions for func in funcs]
        ),
        ufunc_definitions=_indent([*func.define_ufunc.splitlines() for func in funcs]),
```

avalentino (approved):

PR #326, opened 2026-08-26
Test and annotate the ufunc `out`  parameter

eerovaher:
`erfa_generator` automatically translates the ERFA C tests to Python, but that allows only the ERFA functionality to be tested. [The ufunc interface specifies several parameters](https://numpy.org/doc/stable/reference/ufuncs.html#optional-keyword-arguments) common to all ufuncs, but in `pyerfa` they seem to be mostly untested, with the exception of a couple of tests for the `out` parameter. The first commit moves those tests to a separate file and adds a few more tests for `out`. The second commit updates the stub file `erfa_generator` creates so that the `out` parameters are better documented, although the stubs do not specify the dtypes or shapes of any arrays yet. Mypy configuration updates help ensure the stub file is consistent with actual code.

Tests and annotations for the other ufunc interface parameters can be added in the future.


erfa/tests/test_ufunc_interface.py
```diff
@@ -0,0 +1,65 @@
...
+    _, result_with_out, _ = ufunc.cal2jd(*pos_args, out=(None, mjd, None))
+    assert result_with_out is mjd
+    assert_array_equal(mjd, result)
+
+
+def test_out_ellipsis() -> None:
+    pos_args = (2451540, -7342.5)
+    none_result = ufunc.epj(*pos_args, out=None)
+    assert type(none_result) is np.float64
+    ellipsis_result = ufunc.epj(*pos_args, out=...)
```
eerovaher, line 53: The `out=Ellipsis` possibility was implemented in `numpy` 2.3, so the CI jobs that use an older `numpy` fail. For the test run this is not a big problem because it is simple to skip this test if the installed `numpy` version is too old. It is a bigger problem for the stub file because the stub file cannot check which version of `numpy` is installed. A stub file that includes `out=Ellipsis` is erroneously permissive for older `numpy` and a stub file that does not is needlessly restrictive for newer `numpy`. We'll have to err one way or the other and I'll have to think which way is better.

avalentino: I agree that we could just skip the offending test if `numpy` is too old.


We currently support `numpy` 2.2.x (EOL 09/12/2026) and Python 3.10 (EOL October 2026).
Probably we could start dropping support for some of them if it simplifies things.
In particular dropping `numpy` 2.2.x would solve this specific issue if I understand correctly.

Astropy (current master) requires Python >=3.12 and numpy >= 2.0.

Moreover I see that we still support `numpy` 1.x and I'm not sure it is really needed at this point in time.

@eerovaher, @liberfa/pyerfa-maintainers do you think that it make sense to update out minimum version requirement for Python and Numpy?

mhvk: I think our minimums should not be stricter than astropy, so the most we can do, probably relatively soon, is to go to numpy 2.1.

For the particular issue at hand, I think we should error on being slightly erroneously permissive for numpy<2.2.

mhvk (approved):
Looks good. See in-line comment about numpy versions.

eerovaher:
I've turned this PR to a draft for now. I'll come back to this when we've updated our minimum required dependencies.

erfa/tests/test_ufunc_interface.py
eerovaher: The default strategy for tools such as `pip` or `uv` is to install the latest compatible versions of all packages, in which case an environment with a version of `pyerfa` recent enough to include type stubs will also have a recent enough version of `numpy` that supports `out=...`. I have therefore chosen to include it in the stubs, even though I know it is erroneously permissive for several `numpy` versions we still support.

avalentino (approved):

PR #333, opened 2026-09-03
Update OS and Python versions in CI jobs

eerovaher:
The updates help reduce the amount of manual work that has to be done when newer versions become available or support for older versions is dropped.

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -45,13 +45,13 @@ jobs:
 
           - name: Test with development versions of our dependencies
             os: ubuntu-latest
-            python: '3.14'
+            python: '3.x'
             toxenv: test-devdeps
             toxargs: -v
 
           - name: Documentation build
-            os: ubuntu-latest
-            python: 3.x
+            os: ubuntu-22.04
+            python: '3.10'
             toxenv: build_docs
             toxargs: -v
             apt_packages: graphviz
@@ -95,7 +95,6 @@ jobs:
 
           - name: Tests with external liberfa
             os: ubuntu-latest
-            python: '3.10'
             toxenv: test
             toxargs: -v
             apt_packages: python3-venv python3-pip liberfa-dev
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_workflows.yml
```diff
@@ -45,13 +45,13 @@ jobs:
 
           - name: Test with development versions of our dependencies
             os: ubuntu-latest
-            python: '3.14'
+            python: '3.x'
```
eerovaher, line 48: This no longer has to be manually updated when a new Python version becomes available, see https://github.com/actions/setup-python/blob/5fda3b95a4ea91299a34e894583c3862153e4b97/docs/advanced-usage.md?plain=1#L130-L137

.github/workflows/ci_workflows.yml
```diff
@@ -45,13 +45,13 @@ jobs:
...
             toxenv: test-devdeps
             toxargs: -v
 
           - name: Documentation build
-            os: ubuntu-latest
-            python: 3.x
+            os: ubuntu-22.04
+            python: '3.10'
```
eerovaher, line 54: It is sensible to test the documentation build both with the oldest and the latest supported dependencies because then it is reasonable to assume it will also work with all the versions in between. I decided to use this job for the oldest dependencies and the RTD build for the latest so that `.readthedocs.yml` does not have to be edited at all when support for some Python version is dropped.

avalentino (approved):

.github/workflows/ci_workflows.yml
```diff
@@ -95,7 +95,6 @@ jobs:
 
           - name: Tests with external liberfa
             os: ubuntu-latest
-            python: '3.10'
```
eerovaher, line 98: I don't see any reason to use any specific Python version for this job. It is simpler to use whatever Python version the OS comes with. Right now in GitHub workflows `ubuntu-latest` means Ubuntu 24.04 that comes with Python 3.12, but I expect that soon it will be Ubuntu 26.04 that comes with Python 3.14.

PR #335, opened 2026-09-04
Drop support for Python 3.10 and 3.11

eerovaher:
~~[Python 3.10 is very close to its end-of-life](https://devguide.python.org/versions/), so dropping support for it is not controversial.~~

~~The end-of-life for Python 3.11 is still a year away, but [PEP 695 – Type Parameter Syntax](https://peps.python.org/pep-0695/) makes dropping support for it very appealing and supporting it is not required by [SPEC 0](https://scientific-python.org/specs/spec-0000/).~~

EDIT: Dropping support for even just Python 3.10 requires updating `erfa/ufunc.c`. I don't know C, so someone else is going to have to do that.

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -39,7 +39,7 @@ jobs:
 
           - name: Test with oldest supported versions of our dependencies
             os: ubuntu-22.04
-            python: '3.10'
+            python: '3.11'
             toxenv: test-oldestdeps
             toxargs: -v
 
@@ -51,7 +51,7 @@ jobs:
 
           - name: Documentation build
             os: ubuntu-22.04
-            python: '3.10'
+            python: '3.11'
             toxenv: build_docs
             toxargs: -v
             apt_packages: graphviz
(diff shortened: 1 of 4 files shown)


avalentino:
> Dropping support for even just Python 3.10 requires updating ``erfa/ufunc.c``. 

@eerovaher  Sorry I don't get what you say that. Could you lease explain

eerovaher:
I had a quick look at the logs of the failing CI jobs in this PR and I also had a look at the [Python 3.11 C API change log](https://docs.python.org/3.11/whatsnew/3.11.html#whatsnew311-c-api-porting). One problem is that starting from 3.11 `<Python.h>` no longer includes several standard library headers with the limited API, so extensions have to include what they need explicitly (the change log entry points to https://bugs.python.org/issue?@action=redirect&bpo=45434). Another problem I noticed seems to originate from [PEP 670 – Convert macros to functions in the Python C API](https://peps.python.org/pep-0670/) (in particular see the [Avoid the cast in the limited C API version 3.11](https://peps.python.org/pep-0670/#avoid-the-cast-in-the-limited-c-api-version-3-11) subsection). But that's when I stopped investigating because I had confirmed there are problems with the C code and I don't know how to fix those.

avalentino:
OK, thanks. I will give a look.

I was surprised because we started using the limited API only recently (~1year if I remember correctly) and before that we used to build our wheel agains each of the supported Python versions.

Thanks for the clarification

PR #337, opened 2026-09-09
Drop support for Python 3.10

eerovaher:
[Python 3.10 is only a month or so away from its end-of-life](https://devguide.python.org/versions/), so there is no reason to keep supporting it.

Upgrading from Python 3.10 limited C API to Python 3.11 limited C API is not trivial for two reasons. The first is that `Python.h` no longer includes `<stdlib.h>` and `<string.h>` with limited C API (see the [Python 3.11 change log](https://docs.python.org/3.11/whatsnew/3.11.html#whatsnew311-c-api-porting)), so `ufunc.c` needs to include them itself (the former for `malloc()` and `free()` and the latter for `memcpy()`). The second reason is because of [PEP 670 – Convert macros to functions in the Python C API](https://peps.python.org/pep-0670/). Some macros that performed casts themselves were replaced with functions that in the limited C API do not perform casts, so now the casts have to be performed explicitly by `ufunc.c`.

I would also like to drop support for Python 3.11, but my first attempt (#335) ran into problems, so now I think it's better to handle one Python version at a time.

DISCLAIMER: I know neither Python C API in particular nor C in general, I just followed what I think the compiler errors and warnings in #335 suggested.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,8 +26,8 @@ USE_PY_LIMITED_API = not sysconfig.get_config_var("Py_GIL_DISABLED")
 define_macros: list[tuple[str, str | None]] = []
 options = {}
 if USE_PY_LIMITED_API:
-    define_macros.append(("Py_LIMITED_API", "0x30a00f0"))
-    options["bdist_wheel"] = {"py_limited_api": "cp310"}
+    define_macros.append(("Py_LIMITED_API", "0x30b0000"))
+    options["bdist_wheel"] = {"py_limited_api": "cp311"}
 
 
 def get_liberfa_versions(
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -39,7 +39,7 @@ jobs:
 
           - name: Test with oldest supported versions of our dependencies
             os: ubuntu-22.04
-            python: '3.10'
+            python: '3.11'
             toxenv: test-oldestdeps
             toxargs: -v
 
@@ -51,7 +51,7 @@ jobs:
 
           - name: Documentation build
             os: ubuntu-22.04
-            python: '3.10'
+            python: '3.11'
             toxenv: build_docs
             toxargs: -v
             apt_packages: graphviz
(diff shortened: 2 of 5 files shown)


avalentino (approved):
LGTM

setup.py
```diff
@@ -26,8 +26,8 @@
 define_macros: list[tuple[str, str | None]] = []
 options = {}
 if USE_PY_LIMITED_API:
-    define_macros.append(("Py_LIMITED_API", "0x30a00f0"))
-    options["bdist_wheel"] = {"py_limited_api": "cp310"}
+    define_macros.append(("Py_LIMITED_API", "0x30b00f0"))
```
mhvk: This should end in all zeros -- there was probably some reason `0x30a0000` didn't work but `0x30b0000` should be OK.

mhvk (commented):
Looks good to me too, with one small inline comment.

A larger one is to also change the minimum version in `ci_workflows`: the oldest dependencies run , as well as for `external_liberfa` and `build_against_numpy_1` (the latter can probably disappear soon too, but I agree with keeping those changes separate if possible).

setup.py
eerovaher: The [examples used in Python documentation](https://docs.python.org/3.14/c-api/stable.html#c.Py_LIMITED_API) do end with `0000` instead of `00f0`.

eerovaher:
> A larger one is to also change the minimum version in `ci_workflows`: the oldest dependencies run , 

This PR already updates the Python version used in that job.

> as well as for `external_liberfa` 

I am not aware why the `external_liberfa` job should use any specific Python version, so since d9b363a21173265d27ea5bf38f35ea8d34870a6b this job uses whatever Python the OS used for the job comes with. Right now that's Python 3.12, so dropping support for Python 3.10 does not affect that job at all.

> and `build_against_numpy_1`

That job was removed in e7792af145b5b3da2d583c50d5246e31cf422a35.

mhvk (approved):
OK, looks all good now (and apologies for looking at an outdated `main` on my laptop, which led to wrong requests for CI job changes...).

avalentino (approved):

PR #338, opened 2026-09-11
Drop support for 3.11

eerovaher:
[SPEC 0](https://scientific-python.org/specs/spec-0000/) allowed dropping support for Python 3.11 already a year ago and it is no longer supported by the development versions of neither `numpy` (the most important upstream package) nor `astropy` (the most important downstream package). Requiring Python 3.12 is appealing because that is the first Python version that implemented [PEP 695 – Type Parameter Syntax](https://peps.python.org/pep-0695/).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ classifiers = [
     "Topic :: Scientific/Engineering :: Physics",
 ]
 urls = {Homepage = "https://github.com/liberfa/pyerfa"}
-requires-python = ">=3.11"
+requires-python = ">=3.12"
 dependencies = ["numpy>=2.0.0"]
 dynamic = ["version"]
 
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -70,7 +70,6 @@ addopts = "--doctest-rst"
 xfail_strict = true
 
 [tool.cibuildwheel]
-enable = ["pypy"]
```
eerovaher, line 73: [PyPy](https://pypy.org/) currently supports Python 3.11 (and 2.7), so this PR is also dropping support for PyPy. I don't think this is a problem in practice. PyPy is a JIT-compiled implementation of Python, so it can run Python code faster than CPython. However, a Python JIT compiler cannot speed up C extensions, which are very common in scientific programming, so I think for the vast majority of our users PyPy support is not important at all.

avalentino (approved):

PR #340, opened 2026-09-16
Re-enable mypy strict mode

eerovaher:
On current `main`
```
$ mypy
Success: no issues found in 4 source files
```
but
```
$ mypy --strict
   ...
Found 307 errors in 4 files (checked 4 source files)
```
Most of the reported errors are about functions that have not been annotated yet, so in 8f3758d80af19f6a039a80c6bd587ef7adc78028 the strict mode was turned off. However, those errors can also be suppressed using the [`follow_imports`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-follow_imports) setting, which allows strict mode to be re-enabled.


pyproject.toml
```diff
@@ -108,3 +108,9 @@ files = [
     "erfa/tests/test_ufunc.py",
     "erfa/tests/test_ufunc_interface.py",
 ]
+follow_imports = "silent"
+strict = true
+
+[[tool.mypy.overrides]]
+module = ["erfa"]
+implicit_reexport = true
```
eerovaher, line 116: Without this mypy would report
```
erfa/tests/test_ufunc.py:26: error: Module "erfa" does not explicitly export attribute "__version__"  [attr-defined]
Found 1 error in 1 file (checked 4 source files)
```
Forbidding implicit exports is good everywhere except in `__init__.py` files, as long as the `__init__.py` files only contain import statements (and docstrings), so that particular error should be suppressed with this additional configuration.

avalentino (approved):

PR #341, opened 2026-09-17
Add informative annotations to ufunc status codes

eerovaher:
Although on current `main`
```
$ mypy
Success: no issues found in 4 source files
```
the reality is that the `erfa/ufunc.pyi` type stub uses [`Any`](https://docs.python.org/3.12/library/typing.html#the-any-type) as a placeholder a lot, meaning the type annotations are not very useful yet. My current goal is to remove all explicit uses of `Any` from the type stub. This PR takes a step in that direction by adding informative annotations to the ufunc status codes.

`erfa.ufunc` is implemented as a C extension, so static Python type checkers cannot inspect it and they are forced to believe the type stub blindly. If there are any mistakes in the stub then type checkers can report false errors in code using the ufuncs and IDE tab-completion can suggest wrong code. I have therefore added a new `test_ufunc_types.py` file to help ensure the type stub indeed describes runtime behavior. An analogous file will not be needed for the higher-level Python wrappers in `erfa.core` because they are implemented in Python, so static type checkers can ensure inline annotations are consistent with code implementation (assuming of course that `erfa/ufunc.pyi` does not contain mistakes).


erfa/tests/test_ufunc_types.py
```diff
@@ -0,0 +1,21 @@
...
+import numpy as np
+from numpy.typing import NDArray
+
+from erfa import ufunc
+
+
+def test_status_codes_scalar() -> None:
+    _, _, scode = ufunc.eform(0)
+    assert_type(scode, np.intc | NDArray[np.intc])
+    assert isinstance(scode, np.intc)
```
eerovaher, line 14: When I first opened this PR I used `np.int32` instead of `np.intc`, but that did not work on Windows. See e.g. https://github.com/liberfa/pyerfa/actions/runs/35259754083/job/105332211361#step:7:104

avalentino (approved):

